PR #27, opened 2018-11-29
Babel config

marcelomorgado:
refs #22 and #23 - not tried out w/ react-native yet.


marcelomorgado:
@babel/runtime should be installed to can be used by `tasit` react-native demo

packages/tasit-sdk/package.json
```diff
@@ -16,12 +17,21 @@
...
   "scripts": {
-    "test": "echo \"Error: run tests from root\" && exit 1"
+    "prepublish": "npx babel src --out-dir lib --ignore **/*.test.js",
+    "test": "npx mocha src/index.test.js --require @babel/register"
   },
   "bugs": {
     "url": "https://github.com/tasitlabs/tasitsdk/issues"
   },
   "dependencies": {
-    "tasit-account": "^0.0.1"
```
pcowgill: @marcelomorgado So you're thinking we'll reintroduce this dependency once we publish it, right?

packages/tasit-sdk/src/index.test.js
```diff
@@ -0,0 +1,9 @@
+import { Account } from "./index.js";
```
pcowgill: @marcelomorgado I usually leave off the `.js` suffix. Just a style thing though - either way works.

packages/tasit-sdk/lib/index.js
```diff
@@ -0,0 +1,28 @@
+"use strict";
+
+Object.defineProperty(exports, "__esModule", {
+  value: true
+});
+exports.default = exports.Account = void 0;
+
+var _wallet = require("ethers/wallet");
+
+const create = async () => {
```
pcowgill: @marcelomorgado Do you think it matters that we have this as an async function when `Wallet.createRandom` from ethers isn't?

packages/tasit-sdk/src/index.js
```diff
@@ -1,14 +1,22 @@
 "use strict";
 
-// TODO: Add Babel support
-
 import { Wallet } from "ethers/wallet";
 
-export const create = async () => {
+const create = async () => {
```
pcowgill: @marcelomorgado Do you think it matters that we have this as an async function when Wallet.createRandom from ethers isn't?

pcowgill:
@marcelomorgado I'm going to merge this now, but I left a few comments I'd like you to check out for future PRs. Thanks!

packages/tasit-sdk/src/index.js
marcelomorgado: No, not really.

packages/tasit-sdk/package.json
```diff
@@ -16,12 +17,21 @@
     "url": "git+https://github.com/tasitlabs/tasitsdk.git"
   },
   "scripts": {
-    "test": "echo \"Error: run tests from root\" && exit 1"
+    "prepublish": "npx babel src --out-dir lib --ignore **/*.test.js",
```
pcowgill: @marcelomorgado  "As of npm@5, `prepublish` scripts are deprecated. Use `prepare` for build steps and `prepublishOnly` for upload-only."

PR #28, opened 2018-12-01
WIP: React Native compatibility

pcowgill:


marcelomorgado:
I'm happy with it. Seems to me that isn't a big problem the needs to use ethers object.

pcowgill:
Okay - I agree!

PR #35, opened 2018-12-03
Linting setup

marcelomorgado:
#29 
Should I create a npm `prettier` script?


pcowgill:
@marcelomorgado Yeah, `npm run lint` would be a good script to have in the package.json, and then maybe we should make `npm test` also run prettier before running the test suite.

pcowgill:
@marcelomorgado Also let's make a single commit where we commit all the changes to the files that autoformatting introduced.

pcowgill:
@marcelomorgado What text editor are you using? Prettier format on save in VS Code is magical.

marcelomorgado:
Script `npm run lint` createad and `npm run test` updated.

marcelomorgado:
I'm using atom w/ prettier auto-saving.

marcelomorgado:
About formatting existing files, I've not done that yet because this branch is outdated.

packages/tasit-sdk/package.json
```diff
@@ -18,7 +18,8 @@
   },
   "scripts": {
     "prepublish": "npx babel src --out-dir lib --ignore **/*.test.js",
-    "test": "npx mocha src/index.test.js --require @babel/register"
+    "test": "npm run lint && npx mocha src/index.test.js --require @babel/register",
+		"lint": "npx prettier src/* --write"
```
pcowgill:
```suggestion
    "lint": "npx prettier src/* --write"
```

pcowgill:
Ah yeah, cool, Atom configured like that works too!

Yeah, these configs are fine but I'm also open to changes. And if we change anything we should do it relatively soon, but it doesn't need to be right away. We can make a commit that reformats the project along with any changes.

packages/tasit-sdk/package.json
pcowgill: @marcelomorgado Once we make this change we can merge this PR.

Issue #36, opened 2018-12-03
JS files refactoring


marcelomorgado:
https://github.com/tasitlabs/TasitSDK/pull/33#issuecomment-443807534

PR #43, opened 2018-12-04
Renaming several files

marcelomorgado:
Refs #36

### DIFF
--- a/packages/tasit-sdk/src/ethers.test.js
+++ b/packages/tasit-sdk/src/ethers.test.js
@@ -166,7 +166,7 @@ describe("ethers.js - slow test cases", function() {
 		await ropstenProvider.waitForTransaction(tx.hash);
 	});
 
-	it("should broadcast signed tx", async function() {
+	xit("should broadcast signed tx", async function() {
 		rawTx.nonce = await ropstenProvider.getTransactionCount(wallet.address);
 		const signedTx = await wallet.sign(rawTx);
 		const sentTx = await ropstenProvider.sendTransaction(signedTx);


packages/tasit-sdk/src/ethers.test.js
```diff
@@ -0,0 +1,176 @@
...
+		const oldValue = await contract.getValue();
+		const newValue = `I like cats ${randomWallet.mnemonic}`;
+
+		const tx = await contract.setValue(newValue);
+
+		await waitForEvent("ValueChanged", [wallet.address, oldValue, newValue]);
+		await ropstenProvider.waitForTransaction(tx.hash);
+	});
+
+	it("should broadcast signed tx", async function() {
```
pcowgill: Does this one need an xit too? If not, why not?

packages/tasit-sdk/src/tasitSdk.test.js
```diff
@@ -0,0 +1,9 @@
+import { Account } from "./index";
```
pcowgill: For some reason my gut reaction is that I would have done
`import { Account } from "./tasitSdk";`
even though the current approach is actually more directly testing what the user would actually be importing, which makes for an arguably better test. @marcelomorgado what do you think?

packages/tasit-sdk/src/ethers.test.js
marcelomorgado: Sorry, xits are there by mistake. Do you think that is better they are xits by default?

packages/tasit-sdk/src/tasitSdk.test.js
marcelomorgado: I agree with you, I've fixed it on PR https://github.com/tasitlabs/TasitSDK/blob/c1df5e13930835f64c8c5e37a91d2a4b5e58836e/packages/tasit-sdk/src/tasitSdk.test.js#L1

marcelomorgado: I can fix it on this branch either. You call.

packages/tasit-sdk/src/ethers.test.js
pcowgill: I think until we move to ganache and make them run faster, having timeouts enabled for the test suite and keeping these as xit's is the way to go.

packages/tasit-sdk/src/tasitSdk.test.js
pcowgill: It being changed in the other PR is fine with me! Thanks.

packages/tasit-sdk/src/ethers.test.js
marcelomorgado: Done

PR #44, opened 2018-12-04
Moved some features into a smaller package: tasit-account

marcelomorgado:
Refs #39 

- Tasit-account was implemented based on tasit-sdk configurations;
- Unnecessary async at create function was removed (break change);
- Npm scripts `prepublish` was renamed to `prepare`;
- Lerna test script scope removed.

Obs: I've used `npx lerna add tasit-account --scope=tasit-sdk` do add dependecy;
Obs2: This branch was merged from `features/ethers-tests` that PR wasn't merged yet. I've done right?


marcelomorgado:
I'll move ethers dependencies and tests from `taskt-sdk` to `tasit-account`.

pcowgill:
Thanks for noting that this was a breaking change. Since we're still in in 0.y.z, we're fine not bumping a major version. `"Major version zero (0.y.z) is for initial development. Anything may change at any time. The public API should not be considered stable."` Source: semver.org

I like the removal of async for `Account.create`

Thanks for making that npm script deprecation improvement

Good to know about that lerna command you used for this! The scope flag automatically adds the new packages as a dependency for that package?

Yeah, I think in the case where a PR is a new unit of work that's independently mergeable but is gated by an earlier PR that's almost certain to be merged and is gated by review, branching off of that branch for a new PR is the way to go. In general we'll aim for a PR merge velocity such that that's usually not necessary, though - yesterday was a special case. Thanks!

packages/tasit-account/.babelrc
```diff
@@ -0,0 +1,11 @@
+{
```
pcowgill: For the record, I usually use .js config files since I think most projects (like eslint as well) are pushing people in that direction these days. But that's not worth changing until we happen to be adding a new config for something else.

packages/tasit-account/.gitignore
```diff
@@ -0,0 +1 @@
+node_modules/*
```
pcowgill: Would something like `**/node_modules/*` in the root .gitignore work?

marcelomorgado:
> Thanks for noting that this was a breaking change. Since we're still in in 0.y.z, we're fine not bumping a major version. `"Major version zero (0.y.z) is for initial development. Anything may change at any time. The public API should not be considered stable."` Source: semver.org
Thank you for clarification!
> 
> I like the removal of async for `Account.create`
> 
> Thanks for making that npm script deprecation improvement
> 
> Good to know about that lerna command you used for this! The scope flag automatically adds the new packages as a dependency for that package?
Yes! The dependency problem that I've faced before exists when I've tried to import a module from another module subpackage. With packages of same repo `lerna` seems works fine.
> 
> Yeah, I think in the case where a PR is a new unit of work that's independently mergeable but is gated by an earlier PR that's almost certain to be merged and is gated by review, branching off of that branch for a new PR is the way to go. In general we'll aim for a PR merge velocity such that that's usually not necessary, though - yesterday was a special case. Thanks!
Thanks!

packages/tasit-account/src/ethers.test.js
```diff
@@ -0,0 +1,176 @@
...
+	nonce: 0,
+	gasLimit: 21000,
+	gasPrice: ethers.utils.bigNumberify("20000000000"),
+	to: "0x88a5C2d9919e46F883EB62F7b8Dd9d0CC45bc290",
+	value: ethers.utils.parseEther("0.0000001"),
+	data: "0x",
+	chainId: ethers.utils.getNetwork("ropsten").chainId,
+};
+
+describe("ethers.js - pretests", function() {
```
pcowgill: @marcelomorgado Oh, I suppose I should have added the pretests comment in this PR since this got moved. Did you use `git mv` to rename the file? If so, I think the change might have made it through to here.

pcowgill: Maybe once we resolve the merge conflicts this will be working, though

packages/tasit-account/package.json
```diff
@@ -1,27 +1,39 @@
...
-  "scripts": {
-    "test": "echo \"Error: run tests from root\" && exit 1"
-  },
-  "bugs": {
-    "url": "https://github.com/tasitlabs/tasitsdk/issues"
-  },
-  "dependencies": {
-    "ethers": "^4.0.12"
-  }
+	"name": "tasit-account",
```
pcowgill: @marcelomorgado For future reference, I would have put this formatting change in a separate PR with no other changes. That way if anything was change din here, it's sure to show up in the diff. GitLab has this nice feature "Ignore whitespace changes" in the diff on their UI that I hope GitHub adds soon.

packages/tasit-account/lib/tasitAccount.js
```diff
@@ -0,0 +1,29 @@
+"use strict";
```
pcowgill: @marcelomorgado Sometimes I see this type of dir being named `dist` rather than `lib`. How would you feel about that renaming?

packages/tasit-account/src/ethers.test.js
```diff
@@ -0,0 +1,176 @@
...
+// For a detailed explanation of why, see this GitHub issue:
+// https://github.com/tasitlabs/TasitSDK/issues/24#issuecomment-443576993
+describe("ethers.js - unit tests", function() {
+	before("wallet used to sign transactions should have funds", async () => {
+		const balance = ethers.utils.bigNumberify(await wallet.getBalance());
+
+		assert(balance.gt(zero), "wallet balance is zero");
+	});
+
+	it("should sign a raw transaction", async function() {
```
pcowgill: @marcelomorgado I'll note that an ethers test of sending a raw transaction will eventually move to `tasit-act` or whatever we end up calling the reading and writing from the blockchain. That is if we decide to keep raw ethers tests in the repo.

packages/tasit-account/src/ethers.test.js
pcowgill: Oooh, once you merged develop into this branch it did work!

packages/tasit-account/package.json
pcowgill: In particular in this PR, it made it hard to notice the async and prepare changes

packages/tasit-account/.babelrc
marcelomorgado: noted!

packages/tasit-account/.gitignore
marcelomorgado: I think so but I'm not sure, Want you that I test it?

pcowgill: It's not a high priority - don't worry about it. Thanks! Just something to keep in mind for the future in case it becomes a pain to keep adding files like this.

packages/tasit-account/lib/tasitAccount.js
marcelomorgado: Ok, I'll do that.

packages/tasit-account/src/ethers.test.js
marcelomorgado: Ok

packages/tasit-account/.gitignore
marcelomorgado: Done

Issue #40, opened 2018-12-04
Add contract interaction functionaltiy in tasit-act and expose it via tasit-sdk


pcowgill:
A function for sending transactions
A function for broadcasting signed messages
A function for getting data
A function for listening for events

pcowgill:
Or if it's a lot faster, expose it directly via `tasit-sdk` for now

pcowgill:
This package would be a wrapper around these functions with a nice API for developers new to Ethereum https://docs.ethers.io/ethers.js/html/api-contract.html#connecting-to-existing-contracts

pcowgill:
@marcelomorgado I think `tasit-act` should also have `ethers.js` as a dep, and we can use lerna to keep that dep in sync between `tasit-act` and `tasit-account`

pcowgill:
@marcelomorgado The next time we're looking for which tasks to prioritize, I think this one is a strong candidate to add to the list.

pcowgill:
@marcelomorgado I'm going to close this issue and use the new one you created for the rest of this:
https://github.com/tasitlabs/TasitSDK/issues/75

PR #59, opened 2018-12-11
Use ganache for testing

### LINKED ISSUE #38: ethers.js contract tests using ganache
Reducing necessary time to run test cases who need to wait for transactions confirmation.

marcelomorgado:
### Issue link
https://github.com/tasitlabs/TasitSDK/issues/38

### Auto-close the issue?
Closes #38

### Types of changes

### Observations
- tasit-act remains empty (Is it to be done in this PR or another?);
- ethers.test.js still complete at `tasit-account` project (Its splitting is to be done in this PR or another?);
- I'm not sure if the way that I've written scripts chain is the better one, what do you think @pcowgill?
- I've needed downgrade ganache-cli from 6.2.3 to 6.1.8 (https://github.com/ethers-io/ethers.js/issues/275#issuecomment-440427518);
- Contract event testcase remains slow (even using ganache), Will be necessary tryingout another way (e.g.: https://kalis.me/check-events-solidity-smart-contract-test-truffle/)


marcelomorgado:
I'm thinking here, maybe will be better if ganache lives at root repo because not only `tasit-contracts` do need it.

pcowgill:
1) tasit-act emptiness
I think this can come in a separate PR for this issue https://github.com/tasitlabs/TasitSDK/issues/40

2) ethers.test.js splitting
I'm a fan of small PRs, so let's create an issue for splitting these tests.
I've done that here https://github.com/tasitlabs/TasitSDK/issues/63

3) scripts chaining
I'll comment on the scripts chain approach while reviewing the PR itself

4) Downgrading ganache
Sounds good. Good to know. Thanks.

5) Contract event testcase
Which part is the slow part right now?
Rosco's approach from that blog post sounds good.
Here is another useful utility related to this topic https://github.com/OpenZeppelin/openzeppelin-solidity/blob/master/test/helpers/expectEvent.js

6) Ganache in root repo.
Sounds good. We can use lerna hoisting for this https://github.com/lerna/lerna/blob/master/doc/hoist.md
https://github.com/lerna/lerna/blob/master/doc/hoist.md

marcelomorgado:
5. Event listening is slow, we are currently using the same function that ethers.js uses in their tests, I'll check out those alternatives. Thanks!
7. I've forgotten to comment about ENS tests, they use the mainnet (I don't found how to test they using ganache) and maybe should move into an integration test suite.

pcowgill:
If it's simpler we can skip the ENS tests for now.

Down the road if they are inherently long-running, moving them to an integration test suite would make sense. But I'm not sure from first principles why those tests would need to be especially long-running.

pcowgill:
FYI I'm planning to review this PR first thing tomorrow CST!

package.json
```diff
@@ -5,10 +5,15 @@
...
     "@babel/preset-env": "^7.1.6",
     "lerna": "^3.4.3",
-    "prettier": "^1.15.2"
+    "prettier": "^1.15.3",
+    "prettier-plugin-solidity": "^1.0.0-alpha.12"
   },
   "scripts": {
-    "test": "npx lerna run test"
+    "bootstrap": "npx lerna bootstrap",
+    "ganache:start": "npm run ganache:stop && npx lerna run ganache:start --scope tasit-contracts",
```
pcowgill: So the `ganache:start` run by lerna is the npm script in the `tasit-contracts` package, right?

packages/tasit-account/dist/ethers.test.js
```diff
@@ -0,0 +1,136 @@
+"use strict";
```
pcowgill: Having the dist files around does make it more confusing to review PRs. But I don't see a good way around that

packages/tasit-account/src/ethers.test.js
```diff
@@ -16,7 +16,7 @@ const waitForEvent = async (eventName, expected) => {
   });
 };
 
-const ropstenProvider = ethers.getDefaultProvider("ropsten");
+const provider = new ethers.providers.JsonRpcProvider();
```
pcowgill: Where do we give it the ganache config?

Also, many of the account-related tests probably don't even need a provider. Just the ones where we're checking the balance. I wonder if for clarity we should only define the provider farther down in the test suite in a describe block for provider-related tasit-account tests

packages/tasit-account/src/ethers.test.js
```diff
@@ -120,61 +120,64 @@ describe("ethers.js - unit tests", function() {
...
     expect(name).to.equal(ensSample.name);
   });
 
   it("should instantiate a contract object", async function() {
     const contractABI = [
       "event ValueChanged(address indexed author, string oldValue, string newValue)",
-      "constructor(string value)",
-      "function getValue() view returns (string value)",
-      "function setValue(string value)",
+      "constructor(string memory) public",
```
pcowgill: Out of curiosity, what motivated the `memory` addition? The change to `public` makes sense for testing.

package.json
```diff
@@ -5,10 +5,15 @@
...
+    "ganache:stop": "npx lerna run ganache:stop --scope tasit-contracts",
+    "truffle:migrate": "npx lerna run compile --scope tasit-contracts && npx lerna run migrate  --scope tasit-contracts",
+    "test": "npm run bootstrap && npm run ganache:start && npm run truffle:migrate && npx lerna run test && npm run ganache:stop"
```
pcowgill: This flow makes sense reading it here, but it's worth creating a diagram of it too for community members new to this project. I can do that.

packages/tasit-contracts/README.md
```diff
@@ -0,0 +1,3 @@
+# `tasit-sdk`
+
+Most of the details are in the main README for the monorepo [here](https://github.com/tasitlabs/TasitSDK/blob/develop/README.md)
```
pcowgill: Random comment - I could see the `tasit-contracts` package that has a collection of the latest contracts from major projects becoming a popular standalone utility that other developers might use for testing other libaries too.

packages/tasit-contracts/README.md
```diff
@@ -0,0 +1,3 @@
+# `tasit-sdk`
```
pcowgill:
```suggestion
# `tasit-contracts`
```

packages/tasit-contracts/build/contracts/Migrations.json
```diff
@@ -0,0 +1,1498 @@
+{
```
pcowgill: @marcelomorgado Is this `.gitignore` not working? I'm guessing you didn't want this file from the `build` dir here? Or did you?

packages/tasit-contracts/build/contracts/Migrations.json
```diff
@@ -0,0 +1,1498 @@
...
+  "contractName": "Migrations",
+  "abi": [
```
pcowgill: Do we want to parse this file to get the abi that we then pass into ethers? Or is there a cleaner way to achieve that?

packages/tasit-contracts/package.json
```diff
@@ -0,0 +1,34 @@
...
+  "description": "",
+  "keywords": [
+    "ethereum",
+    "dapp",
+    "truffle",
+    "ganache",
+    "solidity"
+  ],
+  "author": "Tasit Labs",
+  "homepage": "https://github.com/tasitlabs/TasitSDK#readme",
```
pcowgill:
```suggestion
  "homepage": "https://github.com/tasitlabs/TasitSDK/tree/develop/packages/tasit-contracts#readme",
```

packages/tasit-contracts/package.json
```diff
@@ -0,0 +1,34 @@
...
+  "license": "MIT",
+  "repository": {
+    "type": "git",
+    "url": "git+https://github.com/tasitlabs/tasitsdk.git"
+  },
+  "scripts": {
+    "ganache:start": "npx ganache-cli -m 'beach swap combine paper music cook electric bullet trust actress liquid asthma' > /dev/null 2>&1 &",
```
pcowgill: What degree of customization is here in the truffle and ganache scripts? Is it straight from the truffle docs? Or was this pulled from a tutorial somewhere? Just curious. Thanks!

packages/tasit-contracts/package.json
```diff
@@ -0,0 +1,34 @@
...
+    "ganache:stop": "pkill -f ganache-cli",
+    "compile": "npx truffle compile > /dev/null 2>&1 &",
```
pcowgill: So are we expecting that users usually run the test suite from the root of this project, which means that truffle can compile the contracts and start up ganache only once and then tasit-contracts and tasit-act use the same singleton?

But then if a dev wants to work only on `tasit-contracts` or only on `tasit-act` (or `tasit-account` if the tests there continue to use a test blockchain), then the test scripts in those packages would also use lerna to start ganache from `tasit-contracts`?

packages/tasit-contracts/test/SimpleStorage.js
```diff
@@ -0,0 +1,15 @@
+const SimpleStorage = artifacts.require("./SimpleStorage.sol");
+
+contract("SimpleStorage", function(accounts) {
+  it("should get the value", async function() {
+    const instance = await SimpleStorage.deployed();
+
+    const value = await instance.getValue();
```
pcowgill: It will probably be a point of confusion for new developers coming to this project that we'll have to get used to the web3.js API for writing truffle tests but then use the ethers.js API and our own API also in JavaScript for testing our own code. So the ethers.js tests in `tasit-act` will look different than the truffle tests in `tasit-contracts` testing the same contract.

pcowgill:
@marcelomorgado The way you wrote the scripts chain makes sense to me. I added a clarifying question and some thoughts on that in my inline comments. I also think the diagram I suggested making in a testing section of the README will help us reason about other ways we might chain that. But I think it's great as it is!

packages/tasit-account/src/ethers.test.js
marcelomorgado: For default, ethers JsonRpcProvider uses 127.0.0.1:8545 (same as ganache-cli).
Ok, I'll instantiate provider only where it's needed.

packages/tasit-account/src/ethers.test.js
marcelomorgado: It's a solc 0.5 requirement (I haven't the chance to read about that, but the contract wasn't compiled without it). I've just realized that I can change solidity version to use 0.4.x (truffle.conf) and then I'll came back to original contract source code. Is it make sense or you want to use 0.5 solc?

package.json
marcelomorgado: Ok, after move ganache to root repo seems to me that will be a bit more clear.

packages/tasit-contracts/README.md
marcelomorgado: Yes, I agree.

packages/tasit-contracts/build/contracts/Migrations.json
marcelomorgado: It's working, I've written it wrong ("./build/*" instead "build/*"). I think it will work after next commit.

packages/tasit-contracts/build/contracts/Migrations.json
marcelomorgado: Maybe the parse will be necessary to interact with more complex contracts, but for now, [human-readable ABI](https://blog.ricmoo.com/human-readable-contract-abis-in-ethers-js-141902f4d917) is working fine.

packages/tasit-contracts/package.json
marcelomorgado: It was the way that I've founded to run ganache in the background and quiet its screen output easily. 
Reading again truffle scripts seems to me that run it in the background is unnecessary. I'll review these scripts.

packages/tasit-contracts/package.json
marcelomorgado: Yes. I've taught to pull up ganache. That way it'll be started from root repo and it'll no more live in tasit-contracts. Same as working now, one test script in mono repo will call and will run:
1) start ganache (TasitSDK);
2) compile and migrate contracts by truffle (tasit-contracts);
3) test all packages;
4) stop ganache.

We can create specific scripts to test packages individually (e.g., test:tasit-account) that will run just what is necessary (in this case, bypass truffle if tasit-account there is no contract interaction and run just tasit-account test suites).

Is it make sense?

packages/tasit-contracts/test/SimpleStorage.js
marcelomorgado: Yes, I agree. If it will be a big problem, we can have no test at this package and use it only for migrations.

packages/tasit-account/src/ethers.test.js
pcowgill: Oh cool, good to know about the defaults.

packages/tasit-account/src/ethers.test.js
pcowgill: Hmm. We might as well move over to 0.5 now to future-proof things a bit. But that context is really helpful.

Down the road I think we'll want to use whichever solc setting the contracts under test are using.

1) The test contracts that we include for testing ethers at a low level should always be the latest solc version supported by ethers.

2) When we use contracts from other projects and run their existing test suites + our own new tests against those contracts using `tasit-act`, we'll want to stick with the current settings (solc version, Solidity version, Truffle version, etc.) of that project for that part of the test suite.

pcowgill: So regarding this section in this PR, I think it can stay as-is. Do you agree?

package.json
pcowgill: I created this issue to track the documentation suggestion. I can take the lead on that. https://github.com/tasitlabs/TasitSDK/issues/64

packages/tasit-contracts/README.md
pcowgill: I created this issue to remind myself to mention that in the README https://github.com/tasitlabs/TasitSDK/issues/65

packages/tasit-contracts/package.json
pcowgill: In your last comment, did you mean `thought` rather than `taught`?

packages/tasit-contracts/test/SimpleStorage.js
```diff
@@ -0,0 +1,15 @@
+const SimpleStorage = artifacts.require("./SimpleStorage.sol");
+
```
pcowgill:
```suggestion
// Note: It will probably be a point of confusion for new developers
// coming to this project that they'll have to get used to the web3.js
// API for writing truffle tests but then use the ethers.js API and our
// own API also in JavaScript for testing our own code. So the
// ethers.js tests in tasit-act will look different than the truffle tests
// in tasit-contracts testing the same contract.
// For that reason, a possible TODO is removing any truffle tests
// other than those we directly add from 3rd-party projects
```

packages/tasit-account/dist/ethers.test.js
pcowgill: @marcelomorgado 
I'm thinking the flow proposed here could work. So we may be able to remove dist files from source control after all and just keep a hash of the latest dist files.
https://twitter.com/paulcowgill/status/1074713318416355328

If you agree, we can create an issue for this.

marcelomorgado: I agree, great solution!

packages/tasit-contracts/build/contracts/Migrations.json
marcelomorgado: Good point. I haven't thought about this yet. We'll need to read the contract ABI, I'm not sure how the better way to do that 1) Reading the build file from `tasit-contracts` or 2) Having static ABI files saved inside the project that will use it.
The second option makes more sense to me, what do you think?

packages/tasit-contracts/package.json
marcelomorgado: I'm not sure about run ganache-cli via truffle. I've never seen this kind of setup. As ganache is used to replace blockchain, I don't know if is the case to start a blockchain from a dev/deploy tool. I've done a quick research about how others projects are automatizing truffle/ganache, and I've seen many of them that are using Docker to start/stop ganache (with very similar logic that we are doing). What do you think?

packages/tasit-contracts/build/contracts/Migrations.json
pcowgill: I think static human-readable ABIs for the projects we support (and ERC standard static ABIs for those we don't explicitly support but implement that interface) is a sensible place to start. That way it will be sure to draw our attention to it when the ABI changes. If it becomes a regular problem to keep that updated we can look into automation at that point.

pcowgill: Here's a related tool to check out for inspiration https://github.com/joincolony/trufflepig

packages/tasit-contracts/package.json
marcelomorgado: I'm doing exactly this right now, I'm working on `tasit-account` tests (event issue) and what I did was ran `npm run ganache:start` from root repo and then `npm run test` on `tasit-account`.

packages/tasit-contracts/package.json
pcowgill: I think I may have been thinking about TestRPC vs ganache-cli as opposed to using truffle vs. ganache. https://github.com/trufflesuite/ganache-cli/blob/master/README.md

I'm going to close this thread out. Thanks!

packages/tasit-contracts/package.json
pcowgill: Ah yeah, great. Thanks for the clarification. The only part we potentially would want to add is making it so that running `npm test` from within `tasit-account` would automatically call `npm run ganache:start` from the root if it wasn't already started.

marcelomorgado: It doesn't work automatically (I've tried to run `npm run ganache:start` inside `tasit-account`). One way that seems that it will work will be if we extract ganache commands to a service-like file script (ganache.js <start|stop>) and then we can run it from sub packages using `"ganache:start": "../../ganache.js start"`.

marcelomorgado:
> I'm working on solving that test case with an event that is slow.

I've figured out what was happening, ethers.js default pooling interval is 4000ms.
https://docs.ethers.io/ethers.js/html/api-providers.html#properties

pcowgill:
> > I'm working on solving that test case with an event that is slow.
> 
> I've figured out what was happening, ethers.js default pooling interval is 4000ms.
> https://docs.ethers.io/ethers.js/html/api-providers.html#properties

Ah that explains it! Nice find.

packages/tasit-contracts/package.json
pcowgill: Ah ok cool! I don't think we need to create an issue for that for now, but we can keep it in mind if we ever experience that as a pain point.

Issue #94, opened 2019-01-09
Solve TasitAction.Contract open threads


marcelomorgado:
**Events behavior/API**
~https://github.com/tasitlabs/TasitSDK/pull/87#discussion_r245855403~ `medium-high priority` (PR https://github.com/tasitlabs/TasitSDK/pull/122)
~https://github.com/tasitlabs/TasitSDK/pull/87#discussion_r245854456~ `low priority` (resolved)
~https://github.com/tasitlabs/TasitSDK/pull/87#discussion_r245855975~ (https://github.com/tasitlabs/TasitSDK/issues/112)
~https://github.com/tasitlabs/TasitSDK/pull/87#discussion_r245853805~ `high priority` (#100)
~https://github.com/tasitlabs/TasitSDK/pull/87#discussion_r245854694~ `high priority` (#100)
~https://github.com/tasitlabs/TasitSDK/pull/87#discussion_r245855603~ `high priority` (#100)
~https://github.com/tasitlabs/TasitSDK/pull/87#discussion_r245857621~ (#96)
~https://github.com/tasitlabs/TasitSDK/pull/87#discussion_r245857814~ `low priority`
~https://github.com/tasitlabs/TasitSDK/pull/87#discussion_r246470897~ (#97)
~https://github.com/tasitlabs/TasitSDK/pull/95#discussion_r246617427~ (PR https://github.com/tasitlabs/TasitSDK/pull/102)
~https://github.com/tasitlabs/TasitSDK/pull/95#discussion_r246618440~ 

**waitForMessage**
~https://github.com/tasitlabs/TasitSDK/pull/87#discussion_r245861154~ (PR https://github.com/tasitlabs/TasitSDK/pull/102)

**Config**
~https://github.com/tasitlabs/TasitSDK/pull/87#discussion_r245862362~ `low priority` (#103)
~https://github.com/tasitlabs/TasitSDK/pull/87#discussion_r245862559~ `low priority`
~https://github.com/tasitlabs/TasitSDK/pull/87#discussion_r246148458~ `low priority` (#103)
 
**Tests**
~https://github.com/tasitlabs/TasitSDK/pull/87#discussion_r245879842~ (PR https://github.com/tasitlabs/TasitSDK/pull/114)
~https://github.com/tasitlabs/TasitSDK/pull/87#discussion_r245880280~ (#113)
~Solve  Non-Deterministic tests~ (PR https://github.com/tasitlabs/TasitSDK/pull/102)
[...]

PR #114, opened 2019-01-15
fakes call count

marcelomorgado:
### Issue link

Refs #94 > https://github.com/tasitlabs/TasitSDK/pull/87#discussion_r245879842

### Types of changes


packages/tasit-action/src/TasitAction.test.js
```diff
@@ -228,7 +228,7 @@ describe("TasitAction.Contract", () => {
 
       await mineBlocks(provider, 15);
 
-      expect(fakeFn.called).to.be.true;
+      expect(fakeFn.callCount).to.be.at.most(7);
```
pcowgill: Wouldn't it be exactly 7? That said, it's good to know about this `at.most` option

marcelomorgado: Related to: https://github.com/tasitlabs/TasitSDK/pull/115#discussion_r248223253
blocks mining is faster than polling.

pcowgill: Makes sense

pcowgill:
```suggestion
      // not exactly 7 because block mining is faster than polling
      expect(fakeFn.callCount).to.be.at.most(7);
```

pcowgill: Want to quickly accept this before we merge?

pcowgill: Oh never mind, I think you did

Issue #112, opened 2019-01-15
One listener per event (PR #102) - open threads


marcelomorgado:
~https://github.com/tasitlabs/TasitSDK/pull/102#discussion_r247585469~ (resolved)
~https://github.com/tasitlabs/TasitSDK/pull/102#discussion_r247583218~ (PR https://github.com/tasitlabs/TasitSDK/pull/136)
~https://github.com/tasitlabs/TasitSDK/pull/102#discussion_r247581707~ (PR https://github.com/tasitlabs/TasitSDK/pull/136)
~https://github.com/tasitlabs/TasitSDK/pull/102#discussion_r247571801~ (PR https://github.com/tasitlabs/TasitSDK/pull/136)
~https://github.com/tasitlabs/TasitSDK/pull/102#discussion_r247575625~ (resolved)
~https://github.com/tasitlabs/TasitSDK/pull/102#discussion_r247586254~ (#113)
~https://github.com/tasitlabs/TasitSDK/pull/102#discussion_r247577857~ (PR https://github.com/tasitlabs/TasitSDK/pull/116)
~https://github.com/tasitlabs/TasitSDK/pull/102#discussion_r247577268~ (PR https://github.com/tasitlabs/TasitSDK/pull/117)
~https://github.com/tasitlabs/TasitSDK/pull/102#discussion_r247578716~ (PR https://github.com/tasitlabs/TasitSDK/pull/117)
~https://github.com/tasitlabs/TasitSDK/pull/102#discussion_r247597059~ (PR https://github.com/tasitlabs/TasitSDK/pull/117)
~https://github.com/tasitlabs/TasitSDK/pull/102#discussion_r247601165~ (PR https://github.com/tasitlabs/TasitSDK/pull/117)
~https://github.com/tasitlabs/TasitSDK/pull/102#discussion_r247598590~ (PR https://github.com/tasitlabs/TasitSDK/pull/117)
~https://github.com/tasitlabs/TasitSDK/pull/102#discussion_r247578907~ (PR https://github.com/tasitlabs/TasitSDK/pull/115)
~https://github.com/tasitlabs/TasitSDK/pull/102#discussion_r247580073~ (extracted to https://github.com/tasitlabs/TasitSDK/issues/133)
~https://github.com/tasitlabs/TasitSDK/pull/102#discussion_r247582916~ (resolved)
~https://github.com/tasitlabs/TasitSDK/pull/102#discussion_r247595053~ (resolved)
[...]

Issue #110, opened 2019-01-15
TasitAction "confirmation" event is firing on already confirmed transactions.


marcelomorgado:
The "confirmation" event should be fired only when tx is confirmed during listener lifetime.

marcelomorgado:
I've looked deep on this and I've mistaken me about this issue.
What the ethers do on pooling is to call `getTransactionReceipt` "inside" `provider.on(txHash)` event. So the event will be fired even the transaction was confirmed a long time ago. The user can check that using `confirmations` param of `receipt` object.

Refs: https://github.com/ethers-io/ethers.js/blob/442553620a5eb466ca20c2f81a8e5685f14b7bcd/src.ts/providers/base-provider.ts#L617

pcowgill:
That's good to know!

marcelomorgado:
Could I close this one?

pcowgill:
Hmm, I think even if ethers.js fires in this case, we shouldn’t have the Sdk fire. It can internally check the receipt object. If you agree, isn’t there still an open issue here?

marcelomorgado:
Hmmm, I agree.
What will be our condition on that? `# of confirmations` or `block timestamp`? I think that #confs is more reliable.

pcowgill:
Oohh, wait, are you saying that that confirmation event that fires when we write the tests in this order is a later confirmation event, not the first one?

I think it's good for confirmation to fire multiple times

I just want to make sure that the 1st confirmation doesn't fire if the user adds the listener after the 1st confirmation happened

I think we should be able to write most tests where we create the listener before sending the transaction.

I'm having trouble picturing what the two alternatives for the conditional are. Could you add in some pseudocode? That might make it clearer for me. Thanks!

marcelomorgado:
Hmm, I'll try to be clearer about what I've got:
I've understood that we should have the same behavior between contract.events and tx.event:
Listening to a contract event, you'll only catch it if they fired inside the window of the listening.

E.g:
```javascript
contract.transfer(from, to, amount);
mineBlocks(10)

// will never catch the event above 
contract.on('Transfer', () => {}) 
```

On the other hand:
```javascript
const tx = contract.transfer(from, to, amount);
mineBlocks(10)

// callback executed and message.data.confirmations = 10
tx.on('confirmation', () => {}) 
```




> I just want to make sure that the 1st confirmation doesn't fire if the user adds the listener after the 1st confirmation happened

[...]

PR #115, opened 2019-01-15
ethers removeListener bug testcase

marcelomorgado:
### Issue link

Refs #112 > https://github.com/tasitlabs/TasitSDK/pull/102#discussion_r247578907 [ethers.removeListener() vs ethers.removeAllListeners()]

### Types of changes


packages/tasit-action/src/ethers.test.js
```diff
@@ -83,6 +84,33 @@ describe("ethers.js", () => {
...
+      };
+      contract.on("ValueChanged", eventListener);
+    });
+
+    // Not working
+    contract.removeAllListeners("ValueChanged");
+    // Working
+    // contract.removeListener("ValueChanged", eventListener);
+
+    expect(eventFakeFn.called).to.be.true;
```
pcowgill: This is called because the listener is triggered even though the transaction was already sent? Should we send the transaction after registering the listener for it to be clearer?

marcelomorgado: A) Yes
I think that it's more clear to me how `ether.js` events listening works internally  (related to https://github.com/tasitlabs/TasitSDK/issues/110). 

B) That's works because new block will be mined after the transaction and then that block will check on next `ethers`'s pool call (because that Promise wait is necessary). If do you want, I can rewrite this to change the order. It won't be worked without rewriting because Promise will be waiting forever.

marcelomorgado:
This issue was solved: https://github.com/ethers-io/ethers.js/issues/391
I've created that test case only to issue creation, do you think that is still necessary?

packages/tasit-action/src/ethers.test.js
marcelomorgado: B) I've made this attempt:
- Waiting for 2 * pollingInterval between blocks mining. Simulating how that's works on real blockchain (pollingInterval < blockTime).
That way I've hopefully with that will be easier to poll routine catch de event and then write something like that:

```javascript
    const sentTx = await contract.setValue(newValue);
    contract.on("ValueChanged", listener);
    // after mining, waiting for enough time to doPoll runs. 
    await mineBlocks(provider, 1); 
```
This worked sometimes only if blocks to mining value was increased.
I think that the poll routine isn't determinist (a lot of promise.then()).
Because of that, using `waitForEvent` functions seems the most reliable way to write contract events tests (Note: As same as ethers.js does in their tests).

pcowgill:
@marcelomorgado I guess it couldn't hurt to have the additional test coverage since it was broken before. No strong opinion though.

packages/tasit-action/src/ethers.test.js
pcowgill: What do you mean by `blocks to mining value`?

pcowgill: Hmm yes, heavily using `waitForEvent` because of that seems reasonable

pcowgill: Would you suggest any changes to the ethers.js poll routine based on this?

pcowgill:
> > This issue was solved: [ethers-io/ethers.js#391](https://github.com/ethers-io/ethers.js/issues/391)
> > I've created that test case only to issue creation, do you think that is still necessary?
> 
> @marcelomorgado I guess it couldn't hurt to have the additional test coverage for the ethers.js lib since it was broken before. No strong opinion though.

But only if rewriting this would be an instructive exercise that would help us with fixing and writing future tests too

packages/tasit-action/src/TasitAction.js
```diff
@@ -40,10 +40,8 @@ class Subscription {
     if (eventName !== "error") {
       const { listener } = event;
 
-      // Note: ethers.removeAllListeners(name) not working!
-      this.#ethersEventEmitter.removeListener(
-        this._toEthersEventName(eventName),
-        listener
+      this.#ethersEventEmitter.removeAllListeners(
```
pcowgill: Oh cool, glad you made this change

packages/tasit-action/src/TasitAction.test.js
```diff
@@ -281,7 +281,7 @@ describe("TasitAction.Contract", () => {
       await mineBlocks(provider, 10);
 
       // TODO: Use fake timer
-      await new Promise(resolve => setTimeout(resolve, 200));
+      await wait(200);
```
pcowgill: Moving this to the test helpers file was a great idea

packages/tasit-action/src/ethers.test.js
```diff
@@ -36,6 +38,10 @@ describe("ethers.js", () => {
 
   afterEach("revert blockchain snapshot", async () => {
     await revertFromSnapshot(provider, testcaseSnaphotId);
+
+    // Note: Without that test suite is broking.
```
pcowgill:
```suggestion
    // Note: Without this the test suite is breaking.
```

packages/tasit-action/src/ethers.test.js
```diff
@@ -36,6 +38,10 @@ describe("ethers.js", () => {
...
+    // It is still unclear why
```
pcowgill: Does it remain unclear after your later comments in this PR? https://github.com/tasitlabs/TasitSDK/pull/115#discussion_r248223253

packages/tasit-action/src/testHelpers/helpers.js
```diff
@@ -1,16 +1,11 @@
...
 import { ethers } from "ethers";
 
-export const waitForEvent = async (contract, eventName, expected) => {
+export const waitForEthersEvent = async (eventEmitter, eventName, callback) => {
   return new Promise(function(resolve, reject) {
-    contract.on(eventName, function() {
-      const args = Array.prototype.slice.call(arguments);
+    eventEmitter.on(eventName, (...args) => {
       const event = args.pop();
-      event.removeListener();
```
pcowgill: I like the changes to this function

pcowgill:
@marcelomorgado Noting that after I merged another PR there are a couple merge conflicts in this PR that will require merging develop into this branch. Thanks!

packages/tasit-action/src/ethers.test.js
marcelomorgado: Yes.

packages/tasit-action/src/ethers.test.js
marcelomorgado: > What do you mean by blocks to mining value?

How many blocks to mine.

> Would you suggest any changes to the ethers.js poll routine based on this?

I've think that maybe a function like `waitForPollFinish` will be useful. Should I create an issue sharing that situation?

marcelomorgado:
> But only if rewriting this would be an instructive exercise that would help us with fixing and writing future tests too

I've worked on that test case and in fact, that helps. :+1:

packages/tasit-action/src/ethers.test.js
pcowgill: > How many blocks to mine

Makes sense now!

> ... issue ...

Yep, I think it’s a good idea to be good contributors and regular providers of feedback to ethers.js

marcelomorgado: Done https://github.com/ethers-io/ethers.js/issues/393

Issue #113, opened 2019-01-15
TasitAction event timeout fix


marcelomorgado:
As it's now, all alive listeners are throwing error event after the timeout. That's wrong.
Which condition should  be true to emit an error?

@pcowgill comments:
> `> TIMEOUT since last confirmation of the tx`
> So we'd want to remove the previously-set timeout and set a new one on each confirmation

> The goal should be for just one error event to come in when this happens.
> I think the time when it should be thrown is if
> current time - last confirmation block time > CONFIRMATION_TIMEOUT

Refs: https://github.com/tasitlabs/TasitSDK/pull/102#discussion_r247586254 and https://github.com/tasitlabs/TasitSDK/pull/87#discussion_r245880280

pcowgill:
Oh, and the timer would be set once on the class for an action subscription. Does that make sense?

marcelomorgado:
> Oh, and the timer would be set once on the class for an action subscription. Does that make sense?

Each event will have its own time. Is it?

pcowgill:
> > Oh, and the timer would be set once on the class for an action subscription. Does that make sense?
> 
> Each event will have its own timer. Is it?

Hmm, I guess we should clarify what type of subscription we're talking about.

For a tx subscription, there's really only one timer that matters - if it's getting confirmed often enough. So that can be a single timer that is reset on each confirmation.

As for events, I'm not sure we even need the ability to set timeouts in this library - that can happen in userland.

marcelomorgado:
Got it.

pcowgill:
I'm closing this since it was addressed in https://github.com/tasitlabs/TasitSDK/pull/122

PR #122, opened 2019-01-16
Event timeout bug fix

marcelomorgado:
### Issue link

Refs https://github.com/tasitlabs/TasitSDK/issues/113

### Types of changes

### Notes

What do you think about the way I've done that @pcowgill ?


pcowgill:
@marcelomorgado Should this be
closes #113 
as opposed to
refs #113 ?

packages/tasit-action/src/TasitAction.js
```diff
@@ -180,6 +182,19 @@ class TransactionSubscription extends Subscription {
           return;
         }
 
+        setTimeout(() => {
+          const currentTime = Date.now();
+          const timedOut =
+            currentTime - this.#lastConfirmationTime >= this.getEventsTimeout();
```
pcowgill: We could consider using moment.js for this, but that's probably overkill

packages/tasit-action/src/TasitAction.js
```diff
@@ -180,6 +182,19 @@ class TransactionSubscription extends Subscription {
           return;
         }
 
+        setTimeout(() => {
```
pcowgill: Hmm, but where's the part where we remove the earlier timeout that was set? Isn't that still missing? I think we might need to store a reference to each timeout that was set?

pcowgill:
@marcelomorgado FYI I just committed a tiny change to fix the merge conflict. I'm 99.9% sure it's what you would have done too, but feel free to take a look.

marcelomorgado:
> @marcelomorgado Should this be
> closes #113
> as opposed to
> refs #113 ?

Yeah, I hope that will close the issue. Shoul I use `closes` even I'm not sure that PR will solve the issue?

packages/tasit-action/src/TasitAction.js
marcelomorgado: Oh yes. You're right.
The code below makes sense?

```javascript
        if (this.#timer) clearTimeout(this.#timer);

        this.#timer = setTimeout(() => {
          const currentTime = Date.now();
          const timedOut =
            currentTime - this.#lastConfirmationTime >= this.getEventsTimeout();

          if (timedOut) {
            this._emitErrorEvent(
              new Error(`Event ${eventName} reached timeout.`),
              eventName
            );
          }
        }, this.getEventsTimeout());
```

Should timer be removed on `off()/usubscribe()` methods either?

packages/tasit-action/src/TasitAction.js
marcelomorgado: It's up to you.

pcowgill: Let’s keep it as-is for now

pcowgill:
> Yeah, I hope that will close the issue. Shoul I use closes even I'm not sure that PR is ready enough to solve the issue?

If you’re 75+% sure, go for it. There’s still PR review to catch it if not

packages/tasit-action/src/TasitAction.js
pcowgill: Yes, that code looks good. I think yes, off from confirmations for the transaction and unsubscribe should both clear it

pcowgill:
@marcelomorgado For the merge conflict here, do we want to resolve it like this?

```
<<<<<<< feature/timeout
      await new Promise(resolve =>
        setTimeout(resolve, txSubscription.getEventsTimeout() * 2)
      );
=======
      await wait(200);
>>>>>>> develop
```

```javascript
      await wait(txSubscription.getEventsTimeout() * 2);
```

Or do we need to divide the eventsTimeout by a big integer before multiplying it by 2?

packages/tasit-action/src/TasitAction.js
marcelomorgado: I've realized that the way I've written above the `#timer` field will belong to `TransactionSubscription` class and it isn't right because each timer must belong to their eventListener.
We could keep it that way because now we have only the "confirmation" event. 
But has another issue: Keeping that way it would be necessary to push down `unsubscribe()` and `off()` methods from `Subscription` superclass.
Because these two situations, I've decided to store the timer on eventListener Map and create these two new methods: `Subscription._setEventTimer(eventName, timer)` and `Subscription._clearEventTimerIfExists(eventName)` what do you think about decision @pcowgill ?

Note: I've just committed this change.

packages/tasit-action/src/TasitAction.js
```diff
@@ -172,6 +204,7 @@ class TransactionSubscription extends Subscription {
 
         if (receipt !== null) {
           this.#txConfirmed = true;
```
pcowgill: Is there any code path where we want to set `this.#txConfirmed` back to `false`?

packages/tasit-action/src/TasitAction.js
marcelomorgado: > So if a user calls unsubscribe without clearing event timers manually, are the event timers still cleared?

Yes, that's the idea.

packages/tasit-action/src/TasitAction.js
marcelomorgado: I've changed it to `this.#txConfirmations` ([uncle block PR](https://github.com/tasitlabs/TasitSDK/pull/123)) the `0` is the new `false`.
It's starts w/ 0 and can turnback to 0 when reorg occurs.

packages/tasit-action/src/TasitAction.js
pcowgill: Resolving this

packages/tasit-action/src/TasitAction.js
pcowgill: Ah right. Okay, good.

pcowgill: Resolving this

Issue #104, opened 2019-01-14
TasitAction.Contract reorg and uncle block testcases


marcelomorgado:
Current test case named `should emit error event when orphan/uncle block occurs` isn't simulating an uncle block case.

1. Rename current test case to `should emit error event when **reorg** block occurs`
2. Create an uncle test case following @pcowgill instructions below:

> An uncle is a valid block that arrives slightly too late, after a correct block was added to the main chain. So I think we'd need to make a variant of mineBlocks to mine two valid blocks with the same transaction, one right after another, and see what happens when the first one is added to the chain on the local test blockchain and then the uncle is mined.
>
> I think the test we have right now is closer to testing a blockchain reorg than an uncle scenario (although you could imagine a reorg that happens on top of an old uncle), since you're mining 15 blocks, then reverting to a snapshot, then mining 20, or vice versa.
>
> An example of a reorg: https://www.theblockcrypto.com/tiny/ethereum-classic-has-experienced-multiple-100-reorg-attack/

marcelomorgado:
Refs: https://github.com/tasitlabs/TasitSDK/pull/71#discussion_r244824045

marcelomorgado:
More about reorg:
https://medium.com/blockvigil/how-we-deal-with-chain-reorganization-at-ethvigil-5a8c06859c7
https://en.bitcoin.it/wiki/Chain_Reorganization

More about uncle/orphan:
https://www.investopedia.com/terms/u/uncle-block-cryptocurrency.asp

marcelomorgado:
Insert another transaction among these test cases.
Refs: https://github.com/tasitlabs/TasitSDK/pull/123/#discussion_r248407347

marcelomorgado:
Until now we have these two test cases `block excluded` and `tx confirmations decreases`.
Is it the case to create another one: `block excluded and tx became confirmed again`?

pcowgill:
More details about potentially inserting a transaction:
https://github.com/tasitlabs/TasitSDK/pull/123#discussion_r248407347

pcowgill:
> Until now we have these two test cases `block excluded` and `tx confirmations decreases`.
> Is it the case to create another one: `block excluded and tx became confirmed again`?

Hmm, I think this is the case I'm picturing, yes. Like if suddenly the new most difficult chain has the tx but in a different block

PR #123, opened 2019-01-16
reorg/uncle block event and test cases improvement

marcelomorgado:
### Issue link

Refs https://github.com/tasitlabs/TasitSDK/issues/104

### Types of changes

### Notes

Now, we have two test cases:

1) Emit orphan error event when tx block was excluded.
2) Emit orphan error event when tx.confirmation decreases.

I've not found how explicit create two blocks with the same transaction with `ganache-cli`. Do you know how I can do that?


packages/tasit-action/src/TasitAction.js
```diff
@@ -168,22 +168,29 @@ class TransactionSubscription extends Subscription {
...
 
-        if (receipt !== null) {
-          this.#txConfirmed = true;
-        } else {
-          if (this.#txConfirmed)
-            this._emitErrorEvent(
-              new Error(`Your message has been included in an uncle block.`),
-              eventName
-            );
+        const includedInAnUncle =
```
pcowgill: Nice variable to define this clearly

packages/tasit-action/src/TasitAction.js
```diff
@@ -168,22 +168,29 @@ class TransactionSubscription extends Subscription {
...
+          (receipt === null && this.#txConfirmations > 0) ||
+          (receipt !== null && receipt.confirmations < this.#txConfirmations);
```
pcowgill:
```suggestion
          (receipt !== null && receipt.confirmations <= this.#txConfirmations);
```

packages/tasit-action/src/TasitAction.js
```diff
@@ -168,22 +168,29 @@ class TransactionSubscription extends Subscription {
...
 
+        if (includedInAnUncle) {
+          this._emitErrorEvent(
+            new Error(`Your message has been included in an uncle block.`),
```
pcowgill: In the README for this functionality we'll want to clearly document what we expect the users to do in response to an event like this in their UI

packages/tasit-action/src/TasitAction.js
```diff
@@ -216,6 +223,11 @@ class TransactionSubscription extends Subscription {
     const tx = await this.#txPromise;
     await this.#provider.waitForTransaction(tx.hash);
   };
+
+  // For testing purposes
```
pcowgill: Do we already have a thread somewhere tracking what we want to do with functions that are only used for testing? I think we do, in which case feel free to resolve this one.

packages/tasit-action/src/TasitAction.js
pcowgill: Would `blockReorgOccurred` be a better name than `includedInAnUncle` or no?

packages/tasit-action/src/TasitAction.test.js
```diff
@@ -328,21 +328,66 @@ describe("TasitAction.Contract", () => {
...
+
+        // But asserting fake function, if that throws, test case will fail.
+        errorFn();
+      };
+
+      const snapshotId = await createSnapshot(provider);
+
+      txSubscription = simpleStorage.setValue("hello world");
+
       txSubscription.on("confirmation", confirmationListener);
```
pcowgill: I think whenever we can we'll want to set it up so users can add the confirmation listener before sending the action

packages/tasit-action/src/TasitAction.test.js
```diff
@@ -328,21 +328,66 @@ describe("TasitAction.Contract", () => {
...
+      await mineBlocks(provider, 10);
+
+      expect(confirmationFn.called).to.be.true;
+
+      await revertFromSnapshot(provider, snapshotId);
+
+      await mineBlocks(provider, 10);
+
+      txSubscription.off("confirmation");
+      txSubscription = undefined;
```
pcowgill: This can  happen in a `beforeEach` or `afterEach`

packages/tasit-action/src/TasitAction.test.js
```diff
@@ -353,20 +398,27 @@ describe("TasitAction.Contract", () => {
...
-      await mineBlocks(provider, 15);
+      await mineBlocks(provider, 20);
 
       expect(confirmationFn.called).to.be.true;
 
       await revertFromSnapshot(provider, snapshotId);
 
-      await mineBlocks(provider, 20);
+      // Note: Without that, ethers.provider will keep the same
+      // receipt.confirmations as before snapshot reversion
```
pcowgill: Interesting! Do you think this is expected behavior on their end?

packages/tasit-action/src/TasitAction.test.js
```diff
@@ -353,20 +398,27 @@ describe("TasitAction.Contract", () => {
...
+      txSubscription.refreshProvider();
 
-      // Broadcast same transaction again (simulate reorg)
-      txSubscription = simpleStorage.setValue("hello world");
+      await mineBlocks(provider, 15);
 
       expect(errorFn.called).to.be.true;
```
pcowgill: Does the error function get called on the first new block because the number of confirmations is lower than before?

packages/tasit-action/src/TasitAction.test.js
```diff
@@ -353,20 +398,27 @@ describe("TasitAction.Contract", () => {
...
       await revertFromSnapshot(provider, snapshotId);
 
-      await mineBlocks(provider, 20);
+      // Note: Without that, ethers.provider will keep the same
+      // receipt.confirmations as before snapshot reversion
+      txSubscription.refreshProvider();
 
-      // Broadcast same transaction again (simulate reorg)
-      txSubscription = simpleStorage.setValue("hello world");
+      await mineBlocks(provider, 15);
```
pcowgill: An idea that I think I've already mentioned elsewhere is that the "reorg" would be more clearly a reorg from reading the test code if you inserted another transaction along the way, like a including a fresh tx before the tx we're interested in and then reinserting the one of interest and then mining more blocks.

pcowgill: But this can certainly come in a follow-up issue/PR that's not the highest priority

packages/tasit-action/src/TasitAction.js
marcelomorgado: Yes, better.

marcelomorgado:
> @marcelomorgado Think we should add an optional Notes section to the PR template?

I think so, I'm using a lot.

packages/tasit-action/src/TasitAction.js
marcelomorgado: Do you want that I create an issue about this?

packages/tasit-action/src/TasitAction.js
marcelomorgado: Yes, here: https://github.com/tasitlabs/TasitSDK/pull/102#discussion_r247713587

packages/tasit-action/src/TasitAction.test.js
marcelomorgado: Extracted to: https://github.com/tasitlabs/TasitSDK/issues/125

marcelomorgado:
> No, good question. I’ll look into it, and we can also create an issue about it on the ganache-cli repo. Or maybe it’s not something we actually have to worry about in practice?

I think that I've aroud that for now, if it's became a problem we could create an issue. Make sense?

packages/tasit-action/src/TasitAction.test.js
marcelomorgado: I've added this comment:
```javascript
      // Note: Transaction no longer exists
      // If it isn't unsed, afterEach hook will execute waitForNonceToUpdate forever
      txSubscription.off("confirmation");
      txSubscription = undefined;
```

What do you think?

packages/tasit-action/src/TasitAction.test.js
marcelomorgado: I'm not sure. But anyway I think that could be useful to share that on this issue: https://github.com/ethers-io/ethers.js/issues/385
What do you think?

packages/tasit-action/src/TasitAction.test.js
marcelomorgado: Yes, not always on the first new block because of pollingInterval vs blockTime issue, but the first poll after that  15 new blocks is emitting error event.

pcowgill:
> > No, good question. I’ll look into it, and we can also create an issue about it on the ganache-cli repo. Or maybe it’s not something we actually have to worry about in practice?
> 
> I think that I've aroud that for now, if it's became a problem we could create an issue. Make sense?

Makes sense!

packages/tasit-action/src/TasitAction.test.js
pcowgill: Got it - okay, I'll add that as a note in the code

pcowgill:
```suggestion
      // not always on the first new block because of pollingInterval vs blockTime issue
      // but the first poll after that 15 new blocks is emitting error event
      expect(errorFn.called).to.be.true;
```

packages/tasit-action/src/TasitAction.test.js
pcowgill: Yeah, great idea!

packages/tasit-action/src/TasitAction.test.js
marcelomorgado: I'm not sure if I've understood well.
Should I create some transaction before that 
```javascript
txSubscription = simpleStorage.setValue("hello world");
``` 
Is it?

packages/tasit-action/src/TasitAction.test.js
pcowgill: Yes, that's perfect!

pcowgill: That's not in the code yet though, right?

packages/tasit-action/src/TasitAction.test.js
pcowgill: Perfect - resolving this

packages/tasit-action/src/TasitAction.js
pcowgill: Sure, thanks. You can assign it to me

packages/tasit-action/src/TasitAction.test.js
marcelomorgado: I've just committed.

packages/tasit-action/src/TasitAction.test.js
marcelomorgado: Diff outdated but I've just commited that.

packages/tasit-action/src/TasitAction.js
marcelomorgado: https://github.com/tasitlabs/TasitSDK/issues/129

packages/tasit-action/src/TasitAction.test.js
marcelomorgado: https://github.com/ethers-io/ethers.js/issues/385#issuecomment-455187735

packages/tasit-action/src/TasitAction.test.js
pcowgill: Resolving this

packages/tasit-action/src/TasitAction.test.js
pcowgill: Thanks! Resolving this.

packages/tasit-action/src/TasitAction.test.js
pcowgill: Thanks! Resolving this

packages/tasit-action/src/TasitAction.test.js
pcowgill: Yeah, I'd have to think carefully about whether this would work, but something like:

```
simpleStorage.setValue("hey earth");
await mineBlocks(provider, 1);
txSubscription = simpleStorage.setValue("hello world");
await mineBlocks(provider, 14);
```

packages/tasit-action/src/TasitAction.js
pcowgill: Thanks! Resolving this.

packages/tasit-action/src/TasitAction.js
pcowgill: Resolving this

pcowgill:
@marcelomorgado I'll let you take the lead on resolving the merge conflict now that I merged the other PR. Thanks!

```javascript
<<<<<<< feature/reorg-uncle
        const blockReorgOccurred =
          (receipt === null && this.#txConfirmations > 0) ||
          (receipt !== null && receipt.confirmations <= this.#txConfirmations);
=======
        if (receipt !== null) {
          this.#txConfirmed = true;
          this.#lastConfirmationTime = Date.now();
        } else {
          if (this.#txConfirmed)
            this._emitErrorEvent(
              new Error(`Your message has been included in an uncle block.`),
              eventName
            );
>>>>>>> develop
```

packages/tasit-action/src/TasitAction.test.js
pcowgill: This can of course come in a later PR if you like that idea

marcelomorgado: I've just committed and renamed the test case to a better name.

marcelomorgado: Sorry, I've unchanged that because introduced a non-deterministic behavior, I'll create a new issue to do that carefully.

marcelomorgado: Appended in already open issue https://github.com/tasitlabs/TasitSDK/issues/104

pcowgill: Great - resolving this

PR #136, opened 2019-01-18
Issue 112 open threads

marcelomorgado:
### Issue link

Refs: #112
https://github.com/tasitlabs/TasitSDK/pull/102#discussion_r247583218 [eventName extra check]
https://github.com/tasitlabs/TasitSDK/pull/102#discussion_r247581707 [destructuring style]
https://github.com/tasitlabs/TasitSDK/pull/102#discussion_r247571801 [rename SimpleStorage contract?]

### Types of changes

### DIFF
--- a/packages/tasit-contracts/test/SimpleStorageWithRemoved.js
+++ b/packages/tasit-contracts/test/SimpleStorageWithRemoved.js
@@ -1,4 +1,4 @@
-const SimpleStorage = artifacts.require("./SimpleStorage.sol");
+const SimpleStorage = artifacts.require("./SimpleStorageWithRemoved.sol");
 // Note: It will probably be a point of confusion for new developers
 // coming to this project that they'll have to get used to the web3.js
 // API for writing truffle tests but then use the ethers.js API and our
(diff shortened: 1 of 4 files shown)


packages/tasit-contracts/test/SimpleStorageWithRemoved.js
```diff
@@ -1,4 +1,4 @@
-const SimpleStorage = artifacts.require("./SimpleStorage.sol");
+const SimpleStorage = artifacts.require("./SimpleStorageWithRemoved.sol");
```
pcowgill: Should we rename the `const` or is that overkill?

marcelomorgado: I've initially thought to rename all vars, but to keep standard I've needed to do that on `tasit-action` test cases either and we'll have a lot of big variables. So I've thought that keep vars with the same name as before a better option. What do you think?

pcowgill: Works for me!

PR #140, opened 2019-01-18
Renaming message to action

### LINKED ISSUE #121: TastAction rename message & transaction to action
We should name as `action` all transactions sent to contract. Code must express that on variables and classes names.

> But I like the goal of finding a word to express contract-changing state (hopefully eventual) transactions. Maybe intendedTransaction? Ooh, or what about action? That way it has the same name as tasit-action.

Refs: https://github.com/tasitlabs/TasitSDK/pull/87#discussion_r245861154

marcelomorgado:
### Auto-close the issue?

Closes #121

### Types of changes

### DIFF
--- a/packages/tasit-action/src/TasitAction.js
+++ b/packages/tasit-action/src/TasitAction.js
@@ -137,6 +137,8 @@ class Subscription {
   };
 }
 
+// Note: If this class will be used to handle with meta-tx too,
+//  that's should be renamed to ActionSubscription.
 class TransactionSubscription extends Subscription {
   #txPromise;
   #provider;
@@ -213,7 +215,7 @@ class TransactionSubscription extends Subscription {
 
         if (blockReorgOccurred) {
           this._emitErrorEvent(
-            new Error(`Your message has been included in an uncle block.`),
+            new Error(`Your action has been included in an uncle block.`),
             eventName
           );
         }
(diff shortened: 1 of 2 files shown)


packages/tasit-action/src/TasitAction.js
```diff
@@ -213,7 +215,7 @@ class TransactionSubscription extends Subscription {
 
         if (blockReorgOccurred) {
           this._emitErrorEvent(
-            new Error(`Your message has been included in an uncle block.`),
+            new Error(`Your action has been included in an uncle block.`),
```
pcowgill:
```suggestion
            new Error(`Your action's position in the chain has changed in a surprising way.`),
```

packages/tasit-action/src/TasitAction.test.js
```diff
@@ -388,7 +388,7 @@ describe("TasitAction.Contract", () => {
 
         // Note: This assertion will not fail the test case (UnhandledPromiseRejectionWarning)
         expect(error.message).to.equal(
-          "Your message has been included in an uncle block."
+          "Your action has been included in an uncle block."
```
pcowgill:
```suggestion
          "Your action's position in the chain has changed in a surprising way."
```

packages/tasit-action/src/TasitAction.test.js
```diff
@@ -436,7 +436,7 @@ describe("TasitAction.Contract", () => {
...
```
pcowgill:
```suggestion
          "Your action's position in the chain has changed in a surprising way."
```

Issue #134, opened 2019-01-17
 Create js middleware (and tests) for ERC721


pcowgill:

marcelomorgado:
TODO:
- More tests
- Extract "mainnet" string hardcoded to const (https://github.com/tasitlabs/TasitSDK/pull/142#discussion_r249165190)

pcowgill:
For the tests, I would focus primarily on the transferring and approval functions first

marcelomorgado:
We are testing those functions already:
`mint`,  `balanceOf`, `transferFrom` and `approve`

pcowgill:
The main other one we might want to test is `safeTransferFrom` https://github.com/OpenZeppelin/openzeppelin-solidity/blob/master/contracts/token/ERC721/IERC721.sol

Down the road maybe something from IERC721Enumerable or IERC721Metadata, but those aren’t urgent. We could create an issue for that that’s a good first issue to test those if you like that idea.

marcelomorgado:
> The main other one we might want to test is `safeTransferFrom` https://github.com/OpenZeppelin/openzeppelin-solidity/blob/master/contracts/token/ERC721/IERC721.sol

OK


> Down the road maybe something from IERC721Enumerable or IERC721Metadata, but those aren’t urgent. We could create an issue for that that’s a `good first issue` to test those if you like that idea.

https://github.com/tasitlabs/TasitSDK/issues/153

pcowgill:
`safeTransferFrom` can also be extracted to another issue and we can close this if you prefer.

marcelomorgado:
I think that could stay here.

marcelomorgado:
I've extracted the `safeTransferFrom` to https://github.com/tasitlabs/TasitSDK/issues/156 issue since it needs that https://github.com/tasitlabs/TasitSDK/issues/155 be solved.

marcelomorgado:
I think that can be closed now.

Issue #90, opened 2019-01-08
tasit-action nonces tracking


marcelomorgado:
Refs: https://github.com/tasitlabs/TasitSDK/pull/87#discussion_r245860818

pcowgill:
In case it gets hard to trace that link above when commits are for old diffs or become resolved, this was the topic: "The thing this function tries to do is waitForTransactionConfirmation (I don't think meta-transactions get confirmed and thus update the nonce automatically? If so we need to address that.)"

marcelomorgado:
Refs: https://github.com/tasitlabs/TasitSDK/pull/147#discussion_r249869326

PR #147, opened 2019-01-22
NFT tests

marcelomorgado:
### Issue link

Refs https://github.com/tasitlabs/TasitSDK/issues/134

### Types of changes

### Notes

- This code has a question about `ethers.BigNumber` vs `BN.js`;
- More test cases should be written?


packages/tasit-action/src/NFT.test.js
```diff
@@ -35,4 +90,69 @@ describe("TasitAction.NFT", () => {
     expect(name).to.exist;
     expect(name).to.equal("Full NFT");
   });
+
+  describe("ERC721 functions", () => {
+    const tokenId = 1;
+
+    beforeEach("should mint one token to owner", async () => {
+      const balanceBefore = await fullNFT.balanceOf(owner.address);
+      expect(toBN(balanceBefore)).to.be.bignumber.equal(toBN(0));
```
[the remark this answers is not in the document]
pcowgill: Like they do here
https://truffleframework.com/docs/truffle/testing/writing-tests-in-javascript#using-async-await

packages/tasit-action/src/NFT.test.js
```diff
@@ -35,4 +90,69 @@ describe("TasitAction.NFT", () => {
...
+        tokenId
+      );
+
+      const event = await new Promise(function(resolve, reject) {
+        contractSubscription.on("Transfer", message => {
+          const { data } = message;
+          const { args } = data;
+          resolve(args);
+        });
+      });
```
pcowgill: We could reject the promise if the event never appears to make the test suite fail quickly when it's not working

packages/tasit-action/src/NFT.test.js
```diff
@@ -35,4 +90,69 @@ describe("TasitAction.NFT", () => {
...
+
+      expect(event.from).to.equal(owner.address);
```
pcowgill: Transfering from ana to bob might be a cleaner example, so the owner transferring is more like airdropping than a normal transfer

packages/tasit-action/src/NFT.test.js
```diff
@@ -35,4 +90,69 @@ describe("TasitAction.NFT", () => {
...
+      expect(event.to).to.equal(ana.address);
+      expect(toBN(event.tokenId)).to.be.bignumber.equal(toBN(tokenId));
+
+      const balanceAfter = await fullNFT.balanceOf(ana.address);
+      expect(toBN(balanceAfter)).to.be.bignumber.equal(toBN(1));
```
pcowgill: Make sure the sender's balance changed too?

packages/tasit-action/src/NFT.test.js
```diff
@@ -35,4 +90,69 @@ describe("TasitAction.NFT", () => {
...
+    });
+
+    it("should transfer an approved token", async () => {
+      txSubscription = fullNFT.approve(ana.address, tokenId);
+
+      await txSubscription.waitForNonceToUpdate();
```
pcowgill: Should we wait for a confirmation event from the `txSubscription` before moving on?

packages/tasit-action/src/NFT.test.js
```diff
@@ -35,4 +90,69 @@ describe("TasitAction.NFT", () => {
...
+
+      fullNFT.setWallet(ana);
```
pcowgill: Would it be cleaner test to make a new fullNFT with ana instead? This is an interesting approach regardless

packages/tasit-action/src/NFT.test.js
```diff
@@ -35,4 +90,69 @@ describe("TasitAction.NFT", () => {
...
+      const balanceBefore = await fullNFT.balanceOf(ana.address);
+      expect(toBN(balanceBefore)).to.be.bignumber.equal(toBN(0));
+
+      txSubscription = fullNFT.transferFrom(
+        owner.address,
+        ana.address,
+        tokenId
+      );
+
+      await txSubscription.waitForNonceToUpdate();
```
pcowgill: Here also, I think it makes sense to wait for a confirmation event explicitly

packages/tasit-action/src/testHelpers/helpers.js
```diff
@@ -40,10 +40,21 @@ export const wait = async ms => {
   return new Promise(resolve => setTimeout(resolve, ms));
 };
 
+// Note: ethers created their own BigNumber type that encapsulates BN.js
+// Because of that, exists the need of extra parses between user and our API (see tests)
+// Should we:
+// - Try to intercept and parse ethers.BigNumber to BN.js (and vice-versa)?
+// or
+// - Exposes ethers.utils.BigNumber / bigNumberify()?
```
pcowgill: I'm leaning towards exposing the ethers.js BigNumber. What functionality does theirs add over BN.js?

It's also worth looking at
https://github.com/trufflesuite/truffle/releases/tag/v5.0.2
and
https://github.com/trufflesuite/truffle/releases/tag/v5.0.0

There's some discussion in how they changed their use of BN in v5, and how it relates to what web3.js v1.0 has done too

pcowgill: In general though, I think we want to embrace and build off of ethers.js's abstractions unless we strongly disagree with the choices they've made

packages/tasit-action/src/NFT.test.js
```diff
@@ -5,15 +5,70 @@ import NFT from "./NFT";
...
+
+      expect(
+        contractSubscription.getEmitter()._events,
+        "ethers.js should not be listening to any events."
+      ).to.be.empty;
+    }
+
+    if (txSubscription) {
+      await txSubscription.waitForNonceToUpdate();
+      txSubscription.unsubscribe();
```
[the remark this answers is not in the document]
marcelomorgado: I don't know, some thoughts:
- `waitForNonceToUpdate` could be called after;
- `waitForNonceToUpdate` will wait for tx promise;
- `waitForNonceToUpdate` is async;
- `unsubscribe` remove all events if exist;
- `unsubscribe` is sync;
- `waitForNonceToUpdate` could not be used if the next `tx` will be sent after the confirmation of the last;

packages/tasit-action/src/NFT.test.js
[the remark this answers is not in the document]
marcelomorgado: I've tried again:

```javascript
      // AssertionError: expected { Object (_hex) } to equal { Object (_hex) }
      //expect(balanceBefore).to.equal(ethers.utils.bigNumberify(0));

      // AssertionError: expected { Object (_hex) } to equal 0
      //expect(balanceBefore).to.equal(0);

      // TypeError: balanceBefore.toBN is not a function
      //expect(balanceBefore.toBN()).to.be.bignumber.equal(0);

      // AssertionError: expected { Object (_hex) } to be an instance of BN
      //expect(balanceBefore).to.be.bignumber.equal(ethers.utils.bigNumberify(0));

      // AssertionError: expected { Object (_hex) } to equal 0
      //expect(balanceBefore.valueOf()).to.equal(0);

      // Working that way but will not work for BN like weis
      //expect(balanceBefore.toNumber()).to.equal(0);

      // Most clear way that works
      expect(balanceBefore.eq(0)).to.be.true;
```

What do you think?

packages/tasit-action/src/NFT.test.js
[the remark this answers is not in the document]
pcowgill: Okay, I'm convinced they should remain separate based on what you said.

It's still not clear to me under which `waitForNonceToUpdate` needs to be called and when it doesn't, though. Have you developed a pretty good sense for that at this point? We can move that discussion elsewhere if appropriate

packages/tasit-action/src/NFT.test.js
```diff
@@ -35,4 +90,69 @@ describe("TasitAction.NFT", () => {
...
+
+  describe("ERC721 functions", () => {
+    const tokenId = 1;
+
+    beforeEach("should mint one token to owner", async () => {
+      const balanceBefore = await fullNFT.balanceOf(owner.address);
+      expect(toBN(balanceBefore)).to.be.bignumber.equal(toBN(0));
+
+      txSubscription = fullNFT.mint(owner.address, tokenId);
+      await txSubscription.waitForNonceToUpdate();
```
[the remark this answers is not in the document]
marcelomorgado: I've attempted to not test so much event here since we have done that on Contract.test.js but I'll follow your suggestion.

packages/tasit-action/src/NFT.test.js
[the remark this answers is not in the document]
pcowgill: I was picturing the second-to-last method in this case.

When the actual number the test is "talking" about is very small, like how many NFTs someone has, I think it makes sense to use native js numbers. When the number is big enough or has a chance of being big enough in the test to warrant using a BN, then do the expectations with big numbers (or I guess the ethers wrapping of BN)

packages/tasit-action/src/NFT.test.js
marcelomorgado: Sure. :+1:

packages/tasit-action/src/NFT.test.js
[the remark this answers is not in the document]
pcowgill: Where does that `eq` function come from?

packages/tasit-action/src/NFT.test.js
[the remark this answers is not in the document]
pcowgill: The reason I bring it up is because I wasn't sure if the nonce having updated would necessarily mean that the transaction was confirmed and therefore that the state had changed, so I thought this could lead to intermittent test failures

packages/tasit-action/src/testHelpers/helpers.js
marcelomorgado: A) Seems that the main motivation is to expose an immutable BN [ricmoo issue comment](https://github.com/ethers-io/ethers.js/issues/58#issuecomment-341993414).

B) Nice! We still using v5-beta, I'll update truffle and see but I think that doesn't affect the way how `ethers` expose numbers.

C) OK, I agree. So, should we expose `TasitAction.Utils` with some BN utils if necessary?

packages/tasit-action/src/NFT.test.js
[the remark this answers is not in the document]
marcelomorgado: I'm using it basically on these situations:
- When the next step of the test case need a warranty that state changed by the last `tx`. (E.g. assert balance after a sent transaction).
- Avoid nonce error.

Note: `waitForNonceToUpdate` also can be changed by "await for event" both will wait until next block/state changed before continuing. But I think that some cases `waitForNonceToUpdate` makes test case more readable and clear (few lines).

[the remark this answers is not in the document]
marcelomorgado: I think that we can extract this thread to an `dicussion` issue if all others became resolved.

packages/tasit-action/src/NFT.test.js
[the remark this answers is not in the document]
marcelomorgado: Rethink about that:
`waitForNonceToUpdate` <=> `waitForTxConfirmation` <=> `waitForNextBlock`
So, after this process wallet nonce is synced with expected nonce for next transaction.

> In practical terms, the nonce is an up-to-date count of the number of confirmed (i.e., on-chain) transactions that have originated from an account. To find out what the nonce is, you can interrogate the blockchain, for example via the web3 interface. Open a JavaScript console in a browser with MetaMask running, or use the truffle console command to access the JavaScript web3 library, then type:

Refs: https://github.com/ethereumbook/ethereumbook/blob/04f66ae45cd9405cce04a088556144be11979699/06transactions.asciidoc#keeping-track-of-nonces

packages/tasit-action/src/NFT.test.js
marcelomorgado: I think that it isn't mandatory

packages/tasit-action/src/NFT.test.js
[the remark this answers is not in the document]
pcowgill: Thanks for that explanation.

Note that the event being awaited here would be the confirmation event from tasit-action txSubscription, not a contract event subscription.

Yeah, extracting this to a discussion issue sounds good.

packages/tasit-action/src/NFT.test.js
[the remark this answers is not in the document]
pcowgill: Ah ok that does help to clear it up. I guess conceivably there could be a block that didn't include the tx, though. But the first two being the same makes sense.

packages/tasit-action/src/NFT.test.js
pcowgill: Same topic as this thread https://github.com/tasitlabs/TasitSDK/pull/147#discussion_r249940935

packages/tasit-action/src/testHelpers/helpers.js
pcowgill: A) Makes sense.
B) Sounds good.
C) Hmm, yeah maybe because an end user would use those utils for writing their own tests? Or would we mostly use them for internal testing and a user of `tasit-action` wouldn't need them?

packages/tasit-action/src/NFT.test.js
```diff
@@ -5,15 +5,77 @@ import NFT from "./NFT";
...
     expect(fullNFT.symbol).to.exist;
     expect(fullNFT.getProvider()).to.exist;
+
+    provider = fullNFT.getProvider();
+
+    snapshotId = await createSnapshot(provider);
+  });
+
+  afterEach("", async () => {
+    // This line avoid that a new event listener catch an existing event from last the block
```
pcowgill:
```suggestion
    // This line ensures that a new event listener does not catch an existing event from last the block
```

packages/tasit-action/src/NFT.test.js
```diff
@@ -93,66 +100,94 @@ describe("TasitAction.NFT", () => {
...
-      expect(toBN(balanceBefore)).to.be.bignumber.equal(toBN(0));
+    beforeEach("should mint one token to ana", async () => {
+      const balanceBefore = await fullNFT.balanceOf(ana.address);
+      expect(balanceBefore.toNumber()).to.equal(0);
 
-      txSubscription = fullNFT.mint(owner.address, tokenId);
-      await txSubscription.waitForNonceToUpdate();
+      contractSubscription = fullNFT.subscribe();
+
+      txSubscription = fullNFT.mint(ana.address, tokenId);
```
pcowgill: `msg.sender` here is the `owner`, right?

packages/tasit-action/src/NFT.test.js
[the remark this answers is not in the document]
pcowgill: I think we can resolve this based on your latest changes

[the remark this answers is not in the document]
pcowgill: Also good to know about `eq`. Thanks.

packages/tasit-action/src/NFT.test.js
pcowgill: I'm resolving this because you addressed it in your latest changes.

packages/tasit-action/src/NFT.test.js
pcowgill: I'm resolving this because you addressed it in your latest changes

packages/tasit-action/src/NFT.test.js
pcowgill: I'm resolving this because you've now addressed this in your latest changes

packages/tasit-action/src/NFT.test.js
pcowgill: We can resolve this since it's a duplicate topic

packages/tasit-action/src/NFT.test.js
pcowgill: Want to accept this quickly before we merge?

packages/tasit-action/src/NFT.test.js
pcowgill: I'm resolving this since it's a duplicate topic

packages/tasit-action/src/NFT.test.js
marcelomorgado: Done.

packages/tasit-action/src/testHelpers/helpers.js
marcelomorgado: I can't see now how exactly our users will use the SDK but, in my experience, handling with BN, `wei/ethers` conversion, hex data, etc. is better with some good utils functions.

packages/tasit-action/src/NFT.test.js
marcelomorgado: yes

packages/tasit-action/src/NFT.test.js
pcowgill: Great! Resolving this

packages/tasit-action/src/NFT.test.js
pcowgill: Okay, good to know. Thanks. Resolving this.

packages/tasit-action/src/NFT.test.js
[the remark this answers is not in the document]
pcowgill: Resolving this

packages/tasit-action/src/testHelpers/helpers.js
pcowgill: Do we want to track this as an issue or should we just keep it in mind? Up to you.

packages/tasit-action/src/NFT.test.js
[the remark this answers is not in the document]
pcowgill: I'll let you create the issue if that's okay. Thanks!

packages/tasit-action/package.json
```diff
@@ -35,8 +35,10 @@
     "@babel/plugin-proposal-class-properties": "^7.2.3",
     "@babel/preset-env": "^7.1.6",
     "@babel/register": "^7.0.0",
+    "bn.js": "^4.11.8",
```
pcowgill: We can remove this in a later PR, I guess, if we're going to use the ethers.js one.

packages/tasit-action/src/testHelpers/helpers.js
marcelomorgado: For me, an issue isn't necessary.

pcowgill: Okay, I'll resolve this then.

packages/tasit-action/src/NFT.test.js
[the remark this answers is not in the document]
marcelomorgado: Do you think that stills as an open thread?

packages/tasit-action/package.json
marcelomorgado: Sure, thanks to remind me :+1:

packages/tasit-action/src/NFT.test.js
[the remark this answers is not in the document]
pcowgill: I guess there might still be an open topic of a new name for this function that makes it clearer when to use it, like `waitForActionReceived`. But we might already have an issue for that. Other than that I think the discussion here is complete and resolved.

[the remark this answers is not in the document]
marcelomorgado: I've linked the issue https://github.com/tasitlabs/TasitSDK/issues/90 to this thread.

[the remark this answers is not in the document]
pcowgill: Perfect, thanks. Resolving this.

Issue #91, opened 2019-01-09
Add tests for tasit-action with Decentraland contracts


pcowgill:

pcowgill:
The additional contract should probably be Decentraland contracts

marcelomorgado:
Project that store the Decentraland contracts: https://github.com/decentraland/land

pcowgill:
And the contracts for the marketplace are here https://github.com/decentraland/marketplace-contracts

marcelomorgado:
Current demo app demands (using NFT class):
- List marketplace lands sell orders (without wallet);
- Get land info (without wallet);
- Buy land.

Issue #135, opened 2019-01-17
Work on the deployment and make sure we know what contracts we’ll point to and on what testnet


pcowgill:

marcelomorgado:
I've found this: https://github.com/decentraland/contracts/blob/gh-pages/addresses.json

pcowgill:
Good find! I wonder what options we have for centralized or decentralized options for finding similar info. I guess the project itself might be the way to go for the most up-to-date info.

Related issue: https://github.com/tasitlabs/TasitSDK/issues/5

marcelomorgado:
We can do that for Decentraland:

`contractAddress` = get from contracts addresses.json
`sourceCodeFromBlockchain` = read solidity source code from etherscan
`sourceCodeFromRepo` = read solidity source conde from our cloned repo

if(`sourceCodeFromBlockchain` == `sourceCodeFromRepo`)
then the `contractAddress` pointing to the same version of the contract that we are syncing

But, if address json is outdated? How to know? Trying to get address from DApp? Trust on their addresses.json?

Ropsten MANA faucet DApp (https://faucet.decentraland.today/)
Mainnet Marketplace DApp (https://market.decentraland.org/marketplace)
All networks addresses: https://contracts.decentraland.org/addresses.json and https://github.com/decentraland/contracts/blob/gh-pages/addresses.json

pcowgill:
For the purposes of this issue, as long we're pretty sure from manual testing that we're pointing to the right one on a testnet (or have deployed our own and are using that address), that's enough.

I think we can move the rest of the discussion [here](
https://github.com/tasitlabs/TasitSDK/issues/9).

These are all very important things to decide on pretty soon - just not before version 0.1.0

marcelomorgado:
A) Since as I know, Ropsten best reproduces the mainnet.
B) I agree.

pcowgill:
@marcelomorgado Hmm, interesting: https://twitter.com/no89thkey/status/1092896560982110208?s=21

marcelomorgado:
Yes, interesting. I've took a look on the thread and seems that the [Gorli](https://goerli.net/) (I didn't know about) testnet will be the widest used one.
But for now, maybe worth keep ropsten because 
- `infura` [doesn't support that yet](https://community.infura.io/t/goerli-testnet-support/305); 
- Decentraland uses ropsten;

pcowgill:
[...]

Issue #9, opened 2018-09-10
Decide on a strategy for staying in sync with 3rd party smart contract versions


pcowgill:
Some initial ideas:

* git submodules in sync with (usually) the master branch
* Etherscan Contract API
* npm for contracts that publish their source there
* APM by Aragon
* EthPM
* It's not ready yet, but in the future I could see something like [pando](https://github.com/ryhope/pando) supporting this

pcowgill:
Here's a related discussion: https://github.com/tasitlabs/TasitSDK/pull/59#discussion_r241790607

pcowgill:
Static human-readable ABIs for the projects we support (and ERC standard static ABIs for those we don't explicitly support but implement that interface) is a sensible place to start. That way it will be sure to draw our attention to it when the ABI changes. If it becomes a regular problem to keep that updated we can look into automation at that point.

Here's a related tool to check out for inspiration in the future https://github.com/joincolony/trufflepig

pcowgill:
Related discussion: https://github.com/tasitlabs/TasitSDK/pull/151#discussion_r250260219

pcowgill:
The solution we're going with for now is to manually update contract versions as needed in [`@tasit/contracts`](https://github.com/tasitlabs/tasit-sdk/tree/develop/packages/contracts) and provide commit hashes for version of the contracts they were based on in the relevant 3rd-party source repo.

If the community converges on a solution, we'll happily move over to that.

PR #151, opened 2019-01-23
Decentraland contracts

marcelomorgado:
### Issue link

Refs #91 

### Types of changes

### Notes

TODO: Test Decentraland contracts using NFT class.

### DIFF
--- a/packages/tasit-action/src/Contract.test.js
+++ b/packages/tasit-action/src/Contract.test.js
@@ -311,7 +311,7 @@ describe("TasitAction.Contract", () => {
       //  https://github.com/sinonjs/sinon/issues/1739
       //  https://github.com/sinonjs/lolex/issues/114
       //  https://stackoverflow.com/a/50785284
-      await wait(txSubscription.getEventsTimeout() * 2);
+      await wait(txSubscription.getEventsTimeout() * 3);
 
       expect(errorFn.called).to.be.true;
       expect(confirmationFn.called).to.be.true;
--- a/packages/tasit-contracts/decentraland/scripts/3_mana_token.js
+++ b/packages/tasit-contracts/decentraland/scripts/3_mana_token.js
@@ -1,0 +1,5 @@
+const MANAToken = artifacts.require("./MANAToken.sol");
+
+module.exports = function(deployer) {
+  deployer.deploy(MANAToken);
+};
--- a/packages/tasit-contracts/decentraland/scripts/install.sh
+++ b/packages/tasit-contracts/decentraland/scripts/install.sh
@@ -1,14 +1,24 @@
 #!/bin/bash
-PROJECT_DIR=$1
-DECENTRALAND_DIR="$PROJECT_DIR/decentraland"
-REPOS="mana land"
+PROJECT_DIR=$1;
+
+if [ -z "$PROJECT_DIR" ];
+then
+    echo "Error! Use: ./$0 <tasit-contract-dir>";
+    exit;
+fi
+
+DECENTRALAND_DIR="$PROJECT_DIR/decentraland";
+REPOS="mana land";
 
 for repo in $REPOS;
 do
-    #if [ ! -d "$DECENTRALAND_DIR/$repo" ]; then
-    rm -rf $DECENTRALAND_DIR/$repo
-    git clone https://github.com/decentraland/$repo.git $DECENTRALAND_DIR/$repo
-    #fi
+    REPO_DIR="$DECENTRALAND_DIR/$repo";
 
-    npm i --prefix $DECENTRALAND_DIR/$repo
+    if [ ! -e "$REPO_DIR/package.json" ];
+    then
+        rm -rf $REPO_DIR;
+        git clone https://github.com/decentraland/$repo.git $REPO_DIR;
+    fi
+
+    npm i --prefix $REPO_DIR;
 done
--- a/packages/tasit-contracts/decentraland/scripts/install.sh
+++ b/packages/tasit-contracts/decentraland/scripts/install.sh
@@ -1,8 +1,12 @@
 #!/bin/bash
 DECENTRALAND_DIR="$PROJECT_DIR/decentraland"
+REPOS="mana land"
 
-if [ ! -d "$DECENTRALAND_DIR/mana" ]; then
-    git clone https://github.com/decentraland/mana.git $DECENTRALAND_DIR/mana
-fi
+for repo in $REPOS;
+do
+    if [ ! -d "$DECENTRALAND_DIR/$repo" ]; then
+        git clone https://github.com/decentraland/$repo.git $DECENTRALAND_DIR/$repo
+    fi
 
-npm i --prefix $DECENTRALAND_DIR/mana
+    npm i --prefix $DECENTRALAND_DIR/$repo
+done
--- a/packages/tasit-contracts/decentraland/scripts/install.sh
+++ b/packages/tasit-contracts/decentraland/scripts/install.sh
@@ -1,0 +1,8 @@
+#!/bin/bash
+DECENTRALAND_DIR="$PROJECT_DIR/decentraland"
+
+if [ ! -d "$DECENTRALAND_DIR/mana" ]; then
+    git clone https://github.com/decentraland/mana.git $DECENTRALAND_DIR/mana
+fi
+
+npm i --prefix $DECENTRALAND_DIR/mana
--- a/packages/tasit-contracts/decentraland/scripts/install.sh
+++ b/packages/tasit-contracts/decentraland/scripts/install.sh
@@ -8,7 +8,7 @@ then
 fi
 
 DECENTRALAND_DIR="$PROJECT_DIR/decentraland";
-REPOS="mana land";
+REPOS="mana land marketplace-contracts";
 
 for repo in $REPOS;
 do
--- a/packages/tasit-contracts/decentraland/scripts/install.sh
+++ b/packages/tasit-contracts/decentraland/scripts/install.sh
@@ -1,4 +1,5 @@
 #!/bin/bash
+PROJECT_DIR=$1
 DECENTRALAND_DIR="$PROJECT_DIR/decentraland"
 REPOS="mana land"
 
--- a/packages/tasit-contracts/decentraland/scripts/install.sh
+++ b/packages/tasit-contracts/decentraland/scripts/install.sh
@@ -5,9 +5,10 @@ REPOS="mana land"
 
 for repo in $REPOS;
 do
-    if [ ! -d "$DECENTRALAND_DIR/$repo" ]; then
-        git clone https://github.com/decentraland/$repo.git $DECENTRALAND_DIR/$repo
-    fi
+    #if [ ! -d "$DECENTRALAND_DIR/$repo" ]; then
+    rm -rf $DECENTRALAND_DIR/$repo
+    git clone https://github.com/decentraland/$repo.git $DECENTRALAND_DIR/$repo
+    #fi
 
     npm i --prefix $DECENTRALAND_DIR/$repo
 done
--- a/packages/tasit-contracts/decentraland/scripts/migrate.sh
+++ b/packages/tasit-contracts/decentraland/scripts/migrate.sh
@@ -2,10 +2,14 @@
 PROJECT_DIR=$1
 DECENTRALAND_DIR="$PROJECT_DIR/decentraland"
 
+echo "##########"
+echo "decentraland dir = $DECENTRALAND_DIR"
+echo "##########"
+
 # mana
 cp $DECENTRALAND_DIR/scripts/3_mana_token.js $DECENTRALAND_DIR/mana/migrations
-cd $DECENTRALAND_DIR/mana && npx truffle migrate
+cd $DECENTRALAND_DIR/mana && ./node_modules/.bin/truffle migrate
 
 # land
 cp $DECENTRALAND_DIR/scripts/3_land_registry.js $DECENTRALAND_DIR/land/migrations
-cd $DECENTRALAND_DIR/land && npx truffle migrate --network livenet
+cd $DECENTRALAND_DIR/land && ./node_modules/.bin/truffle migrate --network livenet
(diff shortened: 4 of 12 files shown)


packages/tasit-contracts/decentraland/scripts/3_mana_token.js
```diff
@@ -0,0 +1,5 @@
+const MANAToken = artifacts.require("./MANAToken.sol");
```
pcowgill: For SQL migrations the rule is that each integer prefix for migration file names should only be used once. Do you think that's the case for Truffle too? I would guess so.

pcowgill: In that case, I think these 3 migrations could move into one file since they all came in one PR

packages/tasit-contracts/decentraland/scripts/install.sh
```diff
@@ -0,0 +1,24 @@
+#!/bin/bash
+PROJECT_DIR=$1;
+
+if [ -z "$PROJECT_DIR" ];
```
pcowgill: Remind me what the `-z` means?

packages/tasit-contracts/decentraland/scripts/install.sh
```diff
@@ -0,0 +1,24 @@
...
+REPOS="mana land marketplace-contracts";
+
+for repo in $REPOS;
+do
+    REPO_DIR="$DECENTRALAND_DIR/$repo";
+
+    if [ ! -e "$REPO_DIR/package.json" ];
+    then
+        rm -rf $REPO_DIR;
+        git clone https://github.com/decentraland/$repo.git $REPO_DIR;
```
pcowgill: This should perhaps change to cloning from a particular release tag or branch

packages/tasit-contracts/decentraland/scripts/migrate.sh
```diff
@@ -0,0 +1,21 @@
...
+
+DECENTRALAND_DIR="$PROJECT_DIR/decentraland"
+
+# mana
+cp $DECENTRALAND_DIR/scripts/3_mana_token.js $DECENTRALAND_DIR/mana/migrations
+cd $DECENTRALAND_DIR/mana && npx truffle migrate
+
+# land
+cp $DECENTRALAND_DIR/scripts/3_land_registry.js $DECENTRALAND_DIR/land/migrations
+cd $DECENTRALAND_DIR/land && npx truffle migrate --network livenet
```
pcowgill: What's the `--network livenet` for?

packages/tasit-contracts/test/LANDRegistry.js
```diff
@@ -0,0 +1,19 @@
+// Note:
+// Should we use ethers instead of web3js?
```
pcowgill: If we're going to use Truffle, I think it makes sense to use their testing framework, which could possibly switch to ethers behind the scenes at some point.

packages/tasit-action/src/Contract.test.js
```diff
@@ -311,7 +311,7 @@ describe("TasitAction.Contract", () => {
       //  https://github.com/sinonjs/sinon/issues/1739
       //  https://github.com/sinonjs/lolex/issues/114
       //  https://stackoverflow.com/a/50785284
-      await wait(txSubscription.getEventsTimeout() * 2);
+      await wait(txSubscription.getEventsTimeout() * 3);
```
pcowgill: Out of curiosity, why did this prove to be useful?

packages/tasit-contracts/decentraland/scripts/install.sh
pcowgill: Do you think this should get merging? That way we would make sure we're testing against the code versions that are deployed as opposed to the bleeding edge of their default GitHub branch.

I'm inclined to say it shouldn't gate merging, but we'll likely want to change this in a follow-on PR.

packages/tasit-action/src/Contract.test.js
marcelomorgado: Yes, that's working. I've raised the value because one of tons tests failed here.

pcowgill:
@marcelomorgado Hmm, is CircleCI not reporting the test results for this PR?

packages/tasit-contracts/decentraland/scripts/migrate.sh
marcelomorgado: `land` project has that network pointing to `8454` port of ganache, the default `develop` network is pointing to `18545` and isn't working.

pcowgill: Ah ok. That might warrant a comment in the code

packages/tasit-contracts/decentraland/scripts/install.sh
marcelomorgado: Yes, you are right. Assuming that the last tag has the last version of deployed contract we should clone from there.
And tracking project/contracts to keep synchronized (refs https://github.com/tasitlabs/TasitSDK/issues/9 and https://github.com/tasitlabs/TasitSDK/issues/135)

packages/tasit-contracts/decentraland/scripts/install.sh
marcelomorgado: > True of the length if "STRING" is zero.

packages/tasit-contracts/decentraland/scripts/3_mana_token.js
marcelomorgado: A) When `truffle migrate` runs that will execute all `js` files inside `migration` following the prefix number order. The `Migrations.last_completed_migration` will store the number/prefix of the last script executed and prevent that already deployed contract be deployed again. When a new contract and a new migration script are created, `truffle migration` will deploy only that new contract.
Note: `truffle migrate --reset` will ignore already deployed contracts and re-redeploy all.

marcelomorgado: B) All of these projects (mana, land and marketplace-contracts) use a shell script for deploy'n'tests, and all have the default `1_initial_migration.js` file and an empty `2_deploy_contracts.js`. The idea was not to change 3rd parties repos, because of that I've created these `3` scripts, one for each repo, to be copied to each repo and allows `truffle migrate` to work.

packages/tasit-contracts/decentraland/scripts/install.sh
pcowgill: Cool. Thanks. Resolving.

packages/tasit-contracts/test/LANDRegistry.js
pcowgill: Resolving this because you gave it a thumbs up. Feel free to delete the note though at some point.

packages/tasit-contracts/decentraland/scripts/install.sh
pcowgill: Okay. We can address this in those issues.

packages/tasit-contracts/decentraland/scripts/3_mana_token.js
pcowgill: A) Makes sense, yes, this is my understanding.

B) Ah ok, that makes sense. Just to consider all possibilities, could we have used their internal deployment scripts without adding our own migration at all? Or would that have been harder? I guess if they weren't using truffle or `ganache-cli` but we wanted to still for that project that could get very complicated quickly, and would warrant our own deployment strategy.

marcelomorgado: I'll paste the script here and create a new comment:

```sh
#!/usr/bin/env bash

# Exit script as soon as a command fails.
set -o errexit

# Executes cleanup function at script exit.
trap cleanup EXIT

cleanup() {
  # Kill the ganache instance that we started (if we started one and if it's still running).
  if [ -n "$ganache_pid" ] && ps -p $ganache_pid > /dev/null; then
    kill -9 $ganache_pid
  fi
}

if [ "$SOLIDITY_COVERAGE" = true ]; then
  ganache_port=8555
else
  ganache_port=8545
fi

ganache_running() {
  nc -z localhost "$ganache_port"
}

start_ganache() {
  # We define 10 accounts with balance 1M ether, needed for high-value tests.
  local accounts=(
    --account="0x2bdd21761a483f71054e14f5b827213567971c676928d9a1808cbfa4b7501200,1000000000000000000000000"
    --account="0x2bdd21761a483f71054e14f5b827213567971c676928d9a1808cbfa4b7501201,1000000000000000000000000"
    --account="0x2bdd21761a483f71054e14f5b827213567971c676928d9a1808cbfa4b7501202,1000000000000000000000000"
    --account="0x2bdd21761a483f71054e14f5b827213567971c676928d9a1808cbfa4b7501203,1000000000000000000000000"
    --account="0x2bdd21761a483f71054e14f5b827213567971c676928d9a1808cbfa4b7501204,1000000000000000000000000"
    --account="0x2bdd21761a483f71054e14f5b827213567971c676928d9a1808cbfa4b7501205,1000000000000000000000000"
    --account="0x2bdd21761a483f71054e14f5b827213567971c676928d9a1808cbfa4b7501206,1000000000000000000000000"
    --account="0x2bdd21761a483f71054e14f5b827213567971c676928d9a1808cbfa4b7501207,1000000000000000000000000"
    --account="0x2bdd21761a483f71054e14f5b827213567971c676928d9a1808cbfa4b7501208,1000000000000000000000000"
    --account="0x2bdd21761a483f71054e14f5b827213567971c676928d9a1808cbfa4b7501209,1000000000000000000000000"
  )

  node_modules/.bin/ganache-cli --gasLimit 0xfffffffffff "${accounts[@]}" --port "$ganache_port" > /dev/null &

  ganache_pid=$!
}

if ganache_running; then
  echo "Using existing ganache instance"
else
  echo "Starting our own ganache instance"
  start_ganache
fi

node_modules/.bin/truffle test "$@"
```

marcelomorgado: > B) Ah ok, that makes sense. Just to consider all possibilities, could we have used their internal deployment scripts without adding our own migration at all? Or would that have been harder? I guess if they weren't using truffle or ganache-cli but we wanted to still for that project that could get very complicated quickly, and would warrant our own deployment strategy.

I've just run the script here with our ganache running and that works. My only note is that has only test script that makes migrate too, will not have a `just migrate` script for these projects. 
But it's doesn't seems a problem, I could:
1) Remove tests from `tasit-contract`
2) Since `tasit-contract` tests runs before `tasit-action` tests these contract will be deployed just in time.
3) Get abi from decentraland projects folders.

Sounds good?

pcowgill: That sounds good to me if that sounds good to you! I'm still open to the earlier approach if you prefer it.

packages/tasit-action/src/Contract.test.js
marcelomorgado: The test failed on expectation `expect(errorFn.called).to.be.true;` so error event wasn't triggered. Since it's a time based event, the only explanation that surges to me is that time waiting wasn't enough. I'm not sure why it happened.

packages/tasit-contracts/decentraland/scripts/3_mana_token.js
marcelomorgado: I think that minimizing touching 3rd parties projects is better. I'll go on that way and see if it'll work.

packages/tasit-action/src/Contract.test.js
pcowgill: Good to know. Thanks for the explanation.

packages/tasit-contracts/decentraland/scripts/3_mana_token.js
marcelomorgado: When `truffle test` is called without migrate (same as the script does above), it deploys the required contracts but only for tests and uses snapshots to clear the blockchain when test ends. Because of that, this script in fact doesn't do the same as `truffle migrate` does. Any idea?

marcelomorgado: Maybe an intermediary solution:
Still pasting migration js files into projects but using their scripts to test them instead of having their tests inside `tasit-contract`.

marcelomorgado: I've thinked a little more and I want to share some points:
- We really need run 3th part intire test suite? It will increate our `lerna test` runtime;
- We could be exposed to 3th part non-deterministic tests (I think that isn't the case of Decentraland);
- I like the idea to have at least one test case for each 3th part contract, for now it's useful to check if the address for each one is correct.

pcowgill: Hmm yeah, so what we really want to run is the local migration (as opposed to testing) script for the 3rd-party contracts. I'll have to look whether enough teams expose a migration script that we can count on that being present (and there's still a risk that they'll use the same defaults and compete with each other for a single ganache instance or something).

I agree that running their whole test suite is probably overkill / risky from a CI perspective except once each time we update their contract versions.

Yeah, one test case for each makes sense. The more the test case is reusable or DRY between different projects, the less work it will be to add it each time.

pcowgill: So given that there's still some more thinking to do here, I think this is okay to merge in its current state. We can track the "how best to run migrations for 3rd-party contracts" in a separate issue.

marcelomorgado: Thanks for the quick response! So, I'll go on with my first approach, alright?

marcelomorgado: OK!

marcelomorgado: https://github.com/tasitlabs/TasitSDK/issues/154

pcowgill: Thanks. Resolving!

pcowgill:
> @marcelomorgado Hmm, is CircleCI not reporting the test results for this PR?

It looks like it is now, so that’s good.

PR #157, opened 2019-01-23
Open threads from last PR

marcelomorgado:
### Auto-close the issue?

Closes threads:
[repos from tag](https://github.com/tasitlabs/TasitSDK/pull/151#discussion_r250260219)
[update truffle](https://github.com/tasitlabs/TasitSDK/pull/147#discussion_r249877923)

### Types of changes

### DIFF
--- a/packages/tasit-contracts/decentraland/scripts/install.sh
+++ b/packages/tasit-contracts/decentraland/scripts/install.sh
@@ -1,0 +1,24 @@
+#!/bin/bash
+PROJECT_DIR=$1;
+
+if [ -z "$PROJECT_DIR" ];
+then
+    echo "Error! Use: ./$0 <tasit-contract-dir>";
+    exit;
+fi
+
+DECENTRALAND_DIR="$PROJECT_DIR/decentraland";
+REPOS="mana land marketplace-contracts";
+
+for repo in $REPOS;
+do
+    REPO_DIR="$DECENTRALAND_DIR/$repo";
+
+    if [ ! -e "$REPO_DIR/package.json" ];
+    then
+        rm -rf $REPO_DIR;
+        git clone https://github.com/decentraland/$repo.git $REPO_DIR;
+    fi
+
+    npm i --prefix $REPO_DIR;
+done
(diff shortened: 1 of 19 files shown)


packages/tasit-contracts/decentraland/scripts/install.sh
```diff
@@ -8,17 +8,43 @@ then
...
+then
+    rm -rf $REPO_DIR;
+    git clone https://github.com/decentraland/$REPO.git $REPO_DIR;
+    npm i --prefix $REPO_DIR;
+fi
+
+
+# land
+REPO="land"
+TAG="deploy/2018-08-31"
```
pcowgill: I know I said tag before, but I'm wondering if the hash of the commit itself would be better, since tags can be changed to point to different commits

pcowgill: Either way this is a big improvement already

pcowgill:
@marcelomorgado Do you agree that it appears CircleCI didn't run for this branch?

marcelomorgado:
> @marcelomorgado Do you agree that it appears CircleCI didn't run for this branch?

It runs on my CircleCI I'm not sure but seems that always occurs on PR creation. I'll do another commit soon and then I think that it will run.

packages/tasit-contracts/decentraland/scripts/install.sh
marcelomorgado: OK, I'll commit this change in a few minutes.

pcowgill:
> It runs on my CircleCI I'm not sure but seems that always occurs on PR creation. I'll do another commit soon and then I think that it will run.

Interesting, yeah it did run. We'll have to look into why that is at some point. Not the top priority though.

packages/tasit-contracts/decentraland/scripts/install.sh
pcowgill: I'm resolving this

marcelomorgado:
> > It runs on my CircleCI I'm not sure but seems that always occurs on PR creation. I'll do another commit soon and then I think that it will run.
> 
> Interesting, yeah it did run. We'll have to look into why that is at some point. Not the top priority though.

https://github.com/tasitlabs/TasitSDK/issues/158

Issue #156, opened 2019-01-23
TasitAction.NFT safeTransferFrom test case 


marcelomorgado:
Refs:
https://github.com/tasitlabs/TasitSDK/issues/134#issuecomment-456867538
https://github.com/OpenZeppelin/openzeppelin-solidity/blob/master/contracts/token/ERC721/IERC721.sol

Note: First, issue https://github.com/tasitlabs/TasitSDK/issues/155 should be solved.

pcowgill:
@marcelomorgado I think we can close this

marcelomorgado:
Yep. Closing.

Issue #155, opened 2019-01-23
TasitAction.Contract transaction reversion handling


marcelomorgado:
We have no expected behavior when `eth_send`/`eth_call` reversion occurs.
Note: I've faced this problem when I was writing `safeTransferFrom` test case and expected an exception when transferring token to a contract that not implement  `onERC721Received`.

Below an example of how `ethers` handles with that situation [refs](https://github.com/ethers-io/ethers.js/issues/368#issuecomment-445041029):

```javascript
contract.someMethod(1, 2, 3).then(
  (result) => {
    // success
  }, 
  (error) => {
    console.log(error);
    // error.reason - The Revert reason; this is what you probably care about. :)
    // Additionally:
    // - error.address - the contract address
    // - error.args - [ BigNumber(1), BigNumber(2), BigNumber(3) ] in this case
    // - error.method - "someMethod()" in this case
    // - error.errorSignature - "Error(string)" (the EIP 838 sighash; supports future custom errors)
    // - error.errorArgs - The arguments passed into the error (more relevant post EIP 838 custom errors)
    // - error.transaction - The call transaction used
});
```

Our code for reference:

```javascript
// draft
txSubscription = contract.toBeRejected();
```

- I think that we can't throw an exception since our contract write-function is sync.
- Make sense that user should register an error listener when instantiating a contract?

Another idea?

pcowgill:
Glad you brought this up. Yeah, this is what I was picturing when I mentioned error events from the contract rather than from other listeners. 

I think for each actionSubscription the user should ideally be able to set up an error listener for that action before sending the action or immediatelt afterwards but before it if processed and rejected by the blockchain.

And yes, being able to configure a catch-all error listener that applies for all actions for that contract by default seems worth adding too.

marcelomorgado:
A) So, both should work?
```javascript
// draft
const s = contract.subscribe();
s.on("error", callback); // catch revert error
txSubscription = contract.toBeRejected();
```
and
```javascript
// draft
txSubscription = contract.toBeRejected();
txSubscription.on("error", callback); // catch revert error
```

Notes: 
- For now `ContractSubscription` trigger error event that occurs inside of ther class;
- Make sense have 0:1  `ContractSubscription` instances per `Contract` (I think that will help to implement this easier).

B) Is that what you are thinking about?

```javascript
// draft
const s = contract.subscribe(onErrorCallback); // as optional function
```

pcowgill:
> So, both should work?

[...]

Issue #149, opened 2019-01-22
TasitAction tests: Reuse hooks


marcelomorgado:
Some of the hooks are duplicated between tests (snapshot logic, wallets assignment, etc.).
Maybe `mochaSetup.js` could be a good place.

pcowgill:
Importing them from a lib of test helper functions would work too, right?

marcelomorgado:
Right.

marcelomorgado:
I agree. :+1:

marcelomorgado:
A fix to be made: Ensure that blockchain will be reverted to the correct snapshot `beforeEach` test case. 
Now, when an exception occurs inside a hook and before `revertFromSnapshot` it's breaking the test suite and demands `ganache-cli` restarting.

marcelomorgado:
Try to keep hooks with max 20 lines.

pcowgill:
@marcelomorgado I think you've made some progress on this issue, right?

marcelomorgado:
Yes, I think that we can close this one and work on remaining issues on #212 e #221

pcowgill:
Okay. Closing this.

PR #159, opened 2019-01-24
Transaction revert error handling

marcelomorgado:
### Issue link

Refs: https://github.com/tasitlabs/TasitSDK/issues/155 and https://github.com/tasitlabs/TasitSDK/issues/156

### Types of changes

### Notes

:warning: Early PR with some questions to discuss.


pcowgill:
@marcelomorgado Cool, thanks! Early PR means `WIP` and `don't merge yet`, right?

marcelomorgado:
> @marcelomorgado Cool, thanks! Early PR means `WIP` and `don't merge yet`, right?

It's ready to merge and it's working without bugs, but has some architectural decisions that could be better before merging.

packages/tasit-action/src/TransactionSubscription.js
```diff
@@ -11,19 +11,46 @@ export class TransactionSubscription extends Subscription {
   #txConfirmations;
   #timeout;
   #lastConfirmationTime;
-
-  constructor(txPromise, provider) {
+  #contract;
+
+  // Contract passing itself to here, enables
+  // TransactionSubscription calling this.#contract._emitErrorEvent
+  // to emit error to the contract catch-all error listener
```
pcowgill: I need to think about this more, but would it be cleaner for the `TransactionSubscription` to emit an error event on the `TransactionSubscription` event emitter that is listened to by the Contract which then emits its own error event relaying that original error payload? So in other words, a contract would always be listening to error events on any event emitters of things that belong to it

packages/tasit-action/src/TransactionSubscription.js
```diff
@@ -11,19 +11,46 @@ export class TransactionSubscription extends Subscription {
...
+  // As that's is the only function called for now, could be better
+  // pass that function as a callback here?
```
pcowgill: I do like this suggestion better than passing in the contract

packages/tasit-action/src/TransactionSubscription.js
```diff
@@ -11,19 +11,46 @@ export class TransactionSubscription extends Subscription {
...
     const { timeout } = events;
 
+    this.#txPromise = txPromise.then(
+      tx => {
+        return tx;
+      },
+      error => {
+        this._emitErrorEvent(
+          new Error(`Action with error: ${error.message}`),
+          null
```
pcowgill: I forget, what's the param that is getting a `null` arg here? `eventName`?

packages/tasit-action/src/TransactionSubscription.js
```diff
@@ -11,19 +11,46 @@ export class TransactionSubscription extends Subscription {
...
+
     this.#timeout = timeout;
-    this.#txPromise = txPromise;
     this.#provider = provider;
     this.#txConfirmations = 0;
+    this.#contract = contract;
+  }
+
+  // TODO: Make protected
+  // Arrow functions in class properties won't be in the prototype and we can't call them with super
```
pcowgill:
```suggestion

  // Arrow functions in class properties won't be in the prototype and we can't call them with super
```

pcowgill: Adding this space because the idea of making it protected and the rationale for it being a non-arrow function are two separate thoughts

packages/tasit-action/src/TransactionSubscription.js
```diff
@@ -11,19 +11,46 @@ export class TransactionSubscription extends Subscription {
...
+  // Refs: https://medium.com/@charpeni/arrow-functions-in-class-properties-might-not-be-as-great-as-we-think-3b3551c440b1
+  _emitErrorEvent(error, eventName) {
+    super._emitErrorEvent(error, eventName);
+    if (this.#contract.subscribedEventNames().includes("error"))
+      this.#contract._emitErrorEvent(error, eventName);
```
pcowgill: I feel like we should be fine to "emit" the error event even if `this.#contract.subscribedEventNames()` doesn't include `error`

packages/tasit-action/src/TransactionSubscription.js
pcowgill: You are using `subscribedEventNames` too though, right? But I have a comment in another thread about why you may not need to use that, in which case the comment above about liking this suggestion better still remains.

packages/tasit-action/src/NFT.test.js
```diff
@@ -189,5 +205,103 @@ describe("TasitAction.NFT", () => {
...
+      expect(contractErrorFakeFn.called).to.be.true;
+
+      senderBalance = await fullNFT.balanceOf(sender);
+      expect(senderBalance.toNumber()).to.equal(1);
+
+      receiverBalance = await fullNFT.balanceOf(receiver);
+      expect(receiverBalance.toNumber()).to.equal(0);
+    });
+
+    it("shouldn't do a safeTransferFrom to a contract that doesn't implement `onERC721Received` - Action Subscription", async () => {
```
pcowgill: It's not clear enough from the name of the test case how this will be different from the test case above ("shouldn't do a safeTransferFrom to a contract that doesn't implement `onERC721Received` - Contract Subscription")

packages/tasit-action/src/NFT.test.js
```diff
@@ -4,6 +4,11 @@ import NFT from "./NFT";
 // See https://github.com/tasitlabs/TasitSDK/issues/138
 const fullNFTAddress = "0x0E86f209729bf54763789CDBcA9E8b94f0FD5333";
 
+// Note: Under the current `tasit-contracts` setup SimpleStorageWithRemoved aways will deployed with this address
+// See https://github.com/tasitlabs/TasitSDK/issues/138
+const simpleStorageWithRemovedAddress =
```
pcowgill:
```suggestion

// Note: This contract is used not because of the particulars of this contract, but
// just because we needed a contract to send an NFT to and have it fail
// when using safeTransferFrom because this contract doesn't implement
// onERC721Received
const simpleStorageWithRemovedAddress =
```

packages/tasit-action/src/NFT.test.js
```diff
@@ -189,5 +205,103 @@ describe("TasitAction.NFT", () => {
...
+    it("shouldn't do a safeTransferFrom to a contract that doesn't implement `onERC721Received` - Action Subscription", async () => {
+      let senderBalance, receiverBalance;
+      const sender = ana.address;
+      const receiver = simpleStorageWithRemovedAddress;
+      const actionErrorFakeFn = sinon.fake();
+
+      fullNFT = new NFT(fullNFTAddress, ana);
+
+      senderBalance = await fullNFT.balanceOf(sender);
+      expect(senderBalance.toNumber()).to.equal(1);
```
pcowgill: We could consider making a helper function maybe that does assertion before for a given sender and receiver in this test file that DRY's these tests up a bit

packages/tasit-action/src/NFT.test.js
pcowgill:
```suggestion
    it("should trigger an error if the user is listening for errors for an action and tries safeTransferFrom to a contract without onERC721Received", async () => {
```

pcowgill: If this looks too verbose, we could always use the same text for the `it` but put a different `describe` above the two similar `it`'s

packages/tasit-action/src/NFT.test.js
```diff
@@ -189,5 +205,103 @@ describe("TasitAction.NFT", () => {
...
+      await txSubscription.waitForNonceToUpdate();
+
+      senderBalance = await fullNFT.balanceOf(ana.address);
+      expect(senderBalance.toNumber()).to.equal(0);
+
+      receiverBalance = await fullNFT.balanceOf(bob.address);
+      expect(receiverBalance.toNumber()).to.equal(1);
+    });
+
+    it("shouldn't do a safeTransferFrom to a contract that doesn't implement `onERC721Received` - Contract Subscription", async () => {
```
pcowgill:
```suggestion
    it("should trigger an error if the user is listening for errors from a contract and tries safeTransferFrom to a contract without onERC721Received", async () => {
```

packages/tasit-action/src/NFT.test.js
```diff
@@ -189,5 +205,103 @@ describe("TasitAction.NFT", () => {
...
+      const actionErrorListener = message => {
+        const { error, eventName } = message;
+        expect(error.message).to.equal(
+          "Action with error: VM Exception while processing transaction: revert"
+        );
+        actionErrorFakeFn();
+      };
+
+      txSubscription = fullNFT.safeTransferFrom(sender, receiver, tokenId);
+      txSubscription.on("error", actionErrorListener);
```
pcowgill: I haven't yet thought of a good way to do this, but it would be cool if we could instantiate the `txSubscription` object and attach the error listener before sending the transaction somehow. But I don't have a concrete suggestion of how to do that in a way that would be a more understandable improvement for the end user.

package.json
```diff
@@ -14,7 +14,7 @@
     "ganache:start": "npx ganache-cli -m 'beach swap combine paper music cook electric bullet trust actress liquid asthma' -l 1000000000 > /dev/null 2>&1 &",
     "ganache:stop": "kill `ps ax|grep ganache-cli|grep -v grep| awk '{print $1}'` 2> /dev/null; exit 0",
     "truffle:migrate": "npx lerna run migrate --scope tasit-contracts",
-    "pretest": "npm run bootstrap && npm run ganache:stop && npm run ganache:start && npm run truffle:migrate",
```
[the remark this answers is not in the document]
marcelomorgado: I've used that only for dev env, to ensure that all installs and links are OK only running `npm test`, but since Decentraland repo the bootstrapping is slower.
We will gain some seconds on CI too because on our config we have the bootstrap step before test step (bootstrap was running twice).

packages/tasit-action/src/Contract.js
```diff
@@ -84,18 +107,23 @@ export class Contract {
...
         throw new Error(`Cannot write data to a Contract without a wallet`);
 
-      const tx = this.#contract[f.name].apply(null, args);
-      const subscription = new TransactionSubscription(tx, this.#provider);
+      const tx = this.#ethersContract[f.name].apply(null, args);
+
+      const subscription = new TransactionSubscription(
+        tx,
+        this.#provider,
+        this
```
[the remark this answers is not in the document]
marcelomorgado: Yes, I've seen that. I've kept the `this.#provider` arg because I'm not sure if we want the function `Contract.getProvider()` (for now it is as "testing purpose function"). Maybe rename to `_getProvider()` ?

packages/tasit-action/src/TransactionSubscription.js
marcelomorgado: Yes, `eventName`

packages/tasit-action/src/Contract.js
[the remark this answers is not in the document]
pcowgill: A) I guess having the option of using different providers for different actions could be nice, so in that way it's nice having it as an explicit param.

B) Maybe we add the ability to do `new TransactionSubscriptionWithPresetProvider(...)` in the future, and that uses this under the hood.

C) Wait, I don't totally follow why the `this.#provider` arg is affected by whether or not we have `Contract.getProvider()`

D) And sure, renaming to the underscore could make sense.

package.json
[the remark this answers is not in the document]
pcowgill: Ah ok, great. Resolving this.

packages/tasit-action/src/TransactionSubscription.js
marcelomorgado: Ok, It'll print `no error listener` warnings.

packages/tasit-action/src/NFT.test.js
marcelomorgado: Brainstorming:
```javascript
// 1
txSubscription = fullNFT.safeTransferFrom(sender, receiver, tokenId).on("error", listener);

// 2
action = fullNFT.safeTransferFrom(sender, receiver, tokenId); 
action.on("error", listener);
acton.send();
```

packages/tasit-action/src/TransactionSubscription.js
pcowgill: That's fine I think

packages/tasit-action/src/TransactionSubscription.js
marcelomorgado: > You are using subscribedEventNames too though, right? 
Oh, yes.

> But I have a comment in another thread about why you may not need to use that, in which case the comment above about liking this suggestion better still remains.
:+1:

packages/tasit-action/src/TransactionSubscription.js
pcowgill: Maybe `_emitErrorEvent` shouldn't assume most errors come from other event listeners by default.

What if we make a new function `_emitErrorFromListenerErrorEvent` that takes an error and an `eventName`, and then decorates the error object itself with the eventName key/value and then emits it using `_emitErrorEvent` under the hood?

packages/tasit-action/src/TransactionSubscription.js
marcelomorgado: Oh, it seems an excellent way to go. I'll do some tryouts.

packages/tasit-action/src/NFT.test.js
pcowgill: Oooooh I love 2!

That way it naturally fits with the meta-tx-first mindset, that a transaction might not necessarily be paid for and sent right away by the user.

pcowgill: And that way, calling write functions on a contract class just build the raw tx object and return it

packages/tasit-action/src/TransactionSubscription.js
pcowgill: Cool! The only part I'd need to think about more is how it works if there are multiple TransactionSubscriptions active

packages/tasit-action/src/NFT.test.js
marcelomorgado: TODO

packages/tasit-action/src/NFT.test.js
marcelomorgado: New issue?

packages/tasit-action/src/TransactionSubscription.js
pcowgill: Yep, I think so. So the contract would just need to be automatically listening for errors on all active `TransactionsSubscription`'s to pass them through to the `Contract` error event.

packages/tasit-action/src/Contract.js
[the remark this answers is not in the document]
pcowgill: Agreed that this may be deprecated based on the other discussion.

C) I get it now. So we either (1) pass just the contract and tx as args and use the get function on the contract to get the provider (and thus use it not just for testing the contract itself). Or (2) we pass just the provider and tx as args, and see if we can make do without passing the contract.

packages/tasit-action/src/NFT.test.js
pcowgill: Sounds good! I'll let you create it. Thanks. Feel free to resolve this once you do.

marcelomorgado: https://github.com/tasitlabs/TasitSDK/issues/162

packages/tasit-action/src/Subscription.js
```diff
@@ -72,7 +72,9 @@ export class Subscription {
   };
 
   // TODO: Make protected
-  _emitErrorEvent = (error, eventName) => {
+  // Arrow functions in class properties won't be in the prototype and we can't call them with super
+  // Refs: https://medium.com/@charpeni/arrow-functions-in-class-properties-might-not-be-as-great-as-we-think-3b3551c440b1
```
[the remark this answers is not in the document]
pcowgill: Resolving this

packages/tasit-action/src/Contract.js
```diff
@@ -7,38 +7,60 @@ import Utils from "./Utils";
 import ProviderFactory from "./ProviderFactory";
 import ContractEventsSubscription from "./ContractEventsSubscription";
 import TransactionSubscription from "./TransactionSubscription";
+import Subscription from "./Subscription";
 
-export class Contract {
+// I'm not sure if extend Subscription is the best approach
+// I'm usually think inheritance as an IS-A relationship
+// I'm not really sure if "Contract is a Subscription" sounds good
```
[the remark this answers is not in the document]
pcowgill: I think we can resolve this for now, though. The comment in the code will remind us to continue to think about it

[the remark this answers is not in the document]
marcelomorgado: OK

packages/tasit-action/src/TransactionSubscription.js
marcelomorgado: When the contract is listening for an error event, and then the user calls 
 `subscribedEventNames()` he gets ["error"] even he didn't registered one:

```javascript
      txSubscription = simpleStorage.setValue("hello world");

      // AssertionError: expected [ 'error' ] to be empty 
      expect(txSubscription.subscribedEventNames()).to.be.empty; 

      txSubscription.on("confirmation", listener);
      txSubscription.off("confirmation");

      // AssertionError: expected [ 'error' ] to be empty
      expect(txSubscription.subscribedEventNames()).to.be.empty; 
```

packages/tasit-action/src/Contract.js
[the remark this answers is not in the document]
marcelomorgado: D) Done
C) Still open

packages/tasit-action/src/Contract.test.js
```diff
@@ -141,7 +141,11 @@ describe("TasitAction.Contract", () => {
     });
   });
 
-  describe("TransactionSubscription - actions (tx) subscriptions tests", async () => {
+  // Should we create some revertFunction at SimpleStorageWithRemoved Contract
+  // and rename it to a short name as SampleContract ?
```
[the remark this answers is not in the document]
marcelomorgado: Done

packages/tasit-action/src/TransactionSubscription.js
marcelomorgado: Still open, I've kept the first approach when I'm waiting your opinion.

packages/tasit-action/src/TransactionSubscription.js
marcelomorgado: I've kept the first approach for now, but I've removed the `subscribedEventNames` call. 
I think that we could close this thread because the open question here is already covered by other threads.

packages/tasit-action/src/TransactionSubscription.js
marcelomorgado: Done

packages/tasit-action/src/TransactionSubscription.js
marcelomorgado: Done

packages/tasit-action/src/NFT.test.js
marcelomorgado: Done

packages/tasit-action/src/TransactionSubscription.js
pcowgill: I suppose we could have that getter function not return error in the cases we don’t want it to. Or we could decide that in a sense error is a subscribed event name here as a result of listening at the contract level. Either is okay with me for this PR. Slight preference for the second option (and in that case just the test expectations would need to change)

packages/tasit-action/src/TransactionSubscription.js
pcowgill: Agreed! Resolving

packages/tasit-action/src/TransactionSubscription.js
pcowgill: Thanks! Resolving.

packages/tasit-action/src/TransactionSubscription.js
pcowgill: Great, thanks. Resolving.

packages/tasit-action/src/Contract.test.js
```diff
@@ -1,16 +1,16 @@
 import Contract from "./Contract";
 import Account from "tasit-account";
 
-// Note: SimpleStorage.json is originally genarated by `tasit-contracts` and was pasted here manually
+// Note: sampleContract.json is originally genarated by `tasit-contracts` and was pasted here manually
```
pcowgill:
```suggestion
// Note: SampleContract.json is originally generated by `tasit-contracts` and was pasted here manually
```

packages/tasit-action/src/Contract.test.js
```diff
@@ -103,77 +103,162 @@ describe("TasitAction.Contract", () => {
...
+      contractSubscription.on("ValueChanged", eventListener);
+
+      txSubscription = sampleContract.setValue("hello world");
+
+      await txSubscription.waitForNonceToUpdate();
+
+      await mineBlocks(provider, 1);
+
+      expect(eventFnFake.callCount).to.equal(1);
+      expect(errorListener.callCount).to.equal(2);
```
pcowgill: Why 2?

packages/tasit-action/src/Contract.test.js
```diff
@@ -103,73 +103,162 @@ describe("TasitAction.Contract", () => {
...
+      const eventListener = () => {
+        eventFnFake();
+        throw new Error();
+      };
+
+      sampleContract.on("error", errorListener);
+
+      contractSubscription = sampleContract.subscribe();
+
+      // Tech debt: Know why changing to .once() eventListener stills be called twice
```
pcowgill:
```suggestion
      // Tech debt: Know why changing to .once() eventListener still is called twice
```

packages/tasit-action/src/Contract.test.js
```diff
@@ -103,73 +103,162 @@ describe("TasitAction.Contract", () => {
...
+
+      txSubscription.on("error", txErrorListener);
+
+      await txSubscription.waitForNonceToUpdate();
+
+      expect(contractErrorListener.callCount).to.equal(1);
+      expect(txErrorListener.callCount).to.equal(1);
+    });
+
+    it("should trigger Contract error event on Event error", async () => {
```
pcowgill:
```suggestion
    it("should trigger Contract error event on Contract Event listener error", async () => {
```

packages/tasit-action/src/Contract.test.js
```diff
@@ -103,73 +103,162 @@ describe("TasitAction.Contract", () => {
...
+
+      await txSubscription.waitForNonceToUpdate();
+
+      await mineBlocks(provider, 1);
+
+      expect(eventFnFake.callCount).to.equal(1);
+      expect(errorListener.callCount).to.equal(2);
+    });
+
+    it("should trigger Contract and ContractEventsSubscription error events on Event error", async () => {
```
pcowgill:
```suggestion
    it("should trigger Contract and ContractEventsSubscription error events on Contract Event listener error", async () => {
```

packages/tasit-action/src/Contract.test.js
```diff
@@ -103,77 +103,162 @@ describe("TasitAction.Contract", () => {
...
+      contractSubscription.on("error", eventErrorListener);
+      contractSubscription.on("ValueChanged", eventListener);
+
+      txSubscription = sampleContract.setValue("hello world");
+
+      await txSubscription.waitForNonceToUpdate();
+
+      await mineBlocks(provider, 1);
+
+      expect(contractErrorListener.callCount).to.equal(2);
```
pcowgill: Why 2?

packages/tasit-action/src/Contract.test.js
```diff
@@ -103,77 +103,162 @@ describe("TasitAction.Contract", () => {
...
+      expect(eventErrorListener.callCount).to.equal(1);
+    });
+
+    // Should contract read functions trigger error event instead throw error?
```
pcowgill: I think events are nice for things that can happen at any time. With the wrong gas, a write might not be added to the blockchain for several blocks, etc. A read, on the other hand, should always work quickly. So I think throwing is more appropriate. It is a bit confusing of a difference though, that we'll have to document well in our docs if we go this route

packages/tasit-action/src/ContractEventsSubscription.js
```diff
@@ -1,15 +1,29 @@
...
 
 class ContractEventsSubscription extends Subscription {
+  #ethersContract;
   #contract;
 
   // Note: We're considering listening multiple events at once
   //    adding eventName, listener params to constructor.
-  constructor(contract) {
-    super(contract);
+  constructor(ethersContract, contract) {
```
pcowgill: Do we have it as a goal to make it not require both of these in the constructor at some point? Like maybe the contract knows about the ContractEventsSubscription instead of vice versa?

For whatever convenience it may offer right now, it's still hard to explain from an design perspective why a `ContractEventsSubscription` would need both

packages/tasit-action/src/ContractEventsSubscription.js
```diff
@@ -1,15 +1,29 @@
...
+    super(ethersContract);
+    this.#ethersContract = ethersContract;
     this.#contract = contract;
   }
 
+  // TODO: Make protected
+  _emitErrorEvent(error, eventName) {
```
pcowgill: I think we can remove `eventName` from this one

packages/tasit-action/src/NFT.test.js
```diff
@@ -43,9 +44,18 @@ describe("TasitAction.NFT", () => {
     expect(fullNFT.getAddress()).to.equal(fullNFTAddress);
     expect(fullNFT.name).to.exist;
     expect(fullNFT.symbol).to.exist;
-    expect(fullNFT.getProvider()).to.exist;
+    expect(fullNFT._getProvider()).to.exist;
+
+    provider = fullNFT._getProvider();
+
+    const ownerBalance = await fullNFT.balanceOf(owner.address);
```
pcowgill: This `beforeEach` is getting a little big. Maybe consider stripping this out into a function that is either called in the `beforeEach` still or just the tests where it's pretty crucial

packages/tasit-action/src/NFT.test.js
```diff
@@ -262,28 +250,24 @@ describe("TasitAction.NFT", () => {
       expect(contractErrorFakeFn.called).to.be.true;
 
       senderBalance = await fullNFT.balanceOf(sender);
-      expect(senderBalance.toNumber()).to.equal(1);
-
       receiverBalance = await fullNFT.balanceOf(receiver);
-      expect(receiverBalance.toNumber()).to.equal(0);
+
+      expect([
```
pcowgill: Probably overkill, but we could make a tiny function called `confirmReceipt(sender, receiver)` that gets the balances and does the expectations

packages/tasit-action/src/TransactionSubscription.js
```diff
@@ -45,12 +42,15 @@ export class TransactionSubscription extends Subscription {
   }
 
   // TODO: Make protected
-  // Arrow functions in class properties won't be in the prototype and we can't call them with super
-  // Refs: https://medium.com/@charpeni/arrow-functions-in-class-properties-might-not-be-as-great-as-we-think-3b3551c440b1
   _emitErrorEvent(error, eventName) {
     super._emitErrorEvent(error, eventName);
-    if (this.#contract.subscribedEventNames().includes("error"))
-      this.#contract._emitErrorEvent(error, eventName);
+    this.#contract._emitErrorEvent(error, eventName);
```
pcowgill: I think `eventName` can be removed here

packages/tasit-action/src/ethers.test.js
```diff
@@ -22,8 +22,12 @@ describe("ethers.js", () => {
...
     expect(wallet.provider).to.be.not.undefined;
 
-    contract = new ethers.Contract(contractAddress, contractABI, wallet);
-    expect(contract.address).to.be.equals(contractAddress);
+    sampleContract = new ethers.Contract(
+      sampleContractAddress,
+      contractABI,
+      wallet
+    );
+    expect(sampleContract.address).to.be.equals(sampleContractAddress);
```
pcowgill: chai expectations should mirror English grammar. "to be equals" doesn't work. Either "to equal" or "to be equal to" would

pcowgill:
```suggestion
    expect(sampleContract.address).to.equal(sampleContractAddress);
```

packages/tasit-contracts/test/SampleContract.js
```diff
@@ -0,0 +1,43 @@
...
+const SampleContract = artifacts.require("./SampleContract.sol");
+// Note: It will probably be a point of confusion for new developers
+// coming to this project that they'll have to get used to the web3.js
+// API for writing truffle tests but then use the ethers.js API and our
+// own API also in JavaScript for testing our own code. So the
+// ethers.js tests in tasit-act will look different than the truffle tests
+// in tasit-contracts testing the same contract.
+// For that reason, a possible TODO is removing any truffle tests
+// other than those we directly add from 3rd-party projects
+contract("SampleContract", function(accounts) {
```
pcowgill:
```suggestion

contract("SampleContract", function(accounts) {
```

packages/tasit-action/src/TransactionSubscription.js
marcelomorgado: Yes, I think so. Maybe we should think more about this error behavior. I'll open a `question` issue with that and another question that I've thought about. Alright?

packages/tasit-action/src/Contract.js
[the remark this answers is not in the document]
marcelomorgado: C) Done

packages/tasit-action/src/Contract.js
```diff
@@ -75,6 +75,16 @@ export class Contract extends Subscription {
...
         this
       );
+
+    //if (!this.#subscription.subscribedEventNames().includes("error")) {
+    const errorListener = message => {
+      const { error } = message;
+      this._emitErrorEvent(new Error(`${error.message}`));
+    };
+
+    this.#subscription.on("error", errorListener);
```
pcowgill: Does the subscription mind if the contract claims its only error listener? We should make a test for what happens if the developer then sets a different error listener explicitly on the subscription after first setting it on the contract

packages/tasit-action/src/Contract.js
```diff
@@ -123,11 +133,15 @@ export class Contract extends Subscription {
...
-        this
-      );
+      const subscription = new TransactionSubscription(tx, this.#provider);
+
+      const errorListener = message => {
+        const { error } = message;
+        this._emitErrorEvent(new Error(`${error.message}`));
+      };
+
+      subscription.on("error", errorListener);
```
pcowgill: The diff would have been more easily readable if we named subscriptions as either `actionSub` or `contractEventsSub` (I end up needed to expand unchanged code to find where I am in the file otherwise). Thanks!

packages/tasit-action/src/Contract.test.js
pcowgill: Oh, I think most of these 2's have changed to 1's in the latest changes

packages/tasit-action/src/Contract.test.js
```diff
@@ -322,40 +411,42 @@ describe("TasitAction.Contract", () => {
...
 
     it("subscription should have one listener per event", async () => {
-      txSubscription = simpleStorage.setValue("hello world");
+      txSubscription = sampleContract.setValue("hello world");
 
       const listener1 = message => {};
       const listener2 = message => {};
 
-      expect(txSubscription.subscribedEventNames()).to.be.empty;
+      expect(txSubscription.subscribedEventNames()).to.not.include(
```
pcowgill:
```suggestion
      expect(txSubscription.subscribedEventNames()).not.to.include(
```

pcowgill: (Just tiny grammar nitpicking)

packages/tasit-action/src/Contract.test.js
```diff
@@ -322,40 +411,42 @@ describe("TasitAction.Contract", () => {
...
     });
 
     it("should remove an event", async () => {
-      txSubscription = simpleStorage.setValue("hello world");
+      txSubscription = sampleContract.setValue("hello world");
 
       const listener1 = message => {};
 
-      expect(txSubscription.subscribedEventNames()).to.be.empty;
+      expect(txSubscription.subscribedEventNames()).to.not.include(
```
pcowgill:
```suggestion
      expect(txSubscription.subscribedEventNames()).not.to.include(
```

packages/tasit-action/src/Contract.test.js
```diff
@@ -322,40 +411,42 @@ describe("TasitAction.Contract", () => {
...
 
-      expect(txSubscription.subscribedEventNames()).to.deep.equal([
-        "confirmation",
-      ]);
+      expect(txSubscription.subscribedEventNames()).to.include("confirmation");
 
       txSubscription.off("confirmation");
 
-      expect(txSubscription.subscribedEventNames()).to.be.empty;
+      expect(txSubscription.subscribedEventNames()).to.not.include(
```
pcowgill:
```suggestion
      expect(txSubscription.subscribedEventNames()).not.to.include(
```

packages/tasit-action/src/Contract.test.js
```diff
@@ -417,35 +416,37 @@ describe("TasitAction.Contract", () => {
...
       txSubscription.on("confirmation", listener1);
 
       expect(() => {
         txSubscription.on("confirmation", listener2);
       }).to.throw();
 
-      expect(txSubscription.subscribedEventNames()).to.deep.equal([
-        "confirmation",
-      ]);
+      expect(txSubscription.subscribedEventNames()).to.include("confirmation");
```
pcowgill: This test seems too flexible. Did you change it to this because it includes "error" as well right now?

packages/tasit-action/src/ContractEventsSubscription.js
```diff
@@ -2,26 +2,12 @@ import Subscription from "./Subscription";
...
   //    adding eventName, listener params to constructor.
-  constructor(ethersContract, contract) {
+  constructor(ethersContract) {
     super(ethersContract);
     this.#ethersContract = ethersContract;
-    this.#contract = contract;
-  }
-
-  // TODO: Make protected
-  _emitErrorEvent(error, eventName) {
```
pcowgill: Cool, glad this moved!

packages/tasit-action/src/Subscription.js
```diff
@@ -84,18 +82,28 @@ export class Subscription {
...
+    let listener = newListener;
+    const errorEventListener = this.#eventListeners.get("error");
+
+    if (errorEventListener) {
+      listener = error => {
+        newListener(error);
+        errorEventListener.listener(error);
+      };
+
+      this.off("error");
```
pcowgill: Why does it do the `off` here?

packages/tasit-action/src/Subscription.js
```diff
@@ -84,18 +82,28 @@ export class Subscription {
...
-  }
+  };
 
   // TODO: Make protected
-  _addErrorListener = listener => {
+  _addErrorListener = newListener => {
+    let listener = newListener;
+    const errorEventListener = this.#eventListeners.get("error");
+
+    if (errorEventListener) {
```
pcowgill: This function could use some code comments to explain why this logic is here. I think it makes sense, but it took me too long to remember just by reading the code

packages/tasit-action/src/Subscription.js
```diff
@@ -72,20 +72,38 @@ export class Subscription {
...
-  _addErrorListener = listener => {
+  _emitErrorEventFromEventListener = (error, eventName) => {
+    error.eventName = eventName;
+    this._emitErrorEvent(error);
+  };
+
+  // TODO: Make protected
+  _addErrorListener = newListener => {
+    let listener = newListener;
+    const errorEventListener = this.#eventListeners.get("error");
```
pcowgill:
```suggestion
    const oldErrorEventListener = this.#eventListeners.get("error");
```

packages/tasit-action/src/Subscription.js
```diff
@@ -84,18 +82,28 @@ export class Subscription {
...
   // TODO: Make protected
-  _addErrorListener = listener => {
+  _addErrorListener = newListener => {
+    let listener = newListener;
+    const errorEventListener = this.#eventListeners.get("error");
+
+    if (errorEventListener) {
+      listener = error => {
+        newListener(error);
+        errorEventListener.listener(error);
```
pcowgill: Oh I get it, is this a hack to allow for two listeners, the catch-call one that the contract added and then the custom one?

packages/tasit-action/src/Contract.js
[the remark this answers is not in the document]
pcowgill: Great! Resolving this.

packages/tasit-action/src/TransactionSubscription.js
pcowgill: Sounds good!

packages/tasit-action/src/Contract.test.js
pcowgill: I'm pretty sure this question is deprecated with the newer changes

packages/tasit-action/src/ContractEventsSubscription.js
pcowgill: But since this is `Outdated` now, maybe it doesn't matter

packages/tasit-action/src/Subscription.js
marcelomorgado: Yes, it's necessary because we are allowing only one listener per event (error included).

packages/tasit-action/src/Contract.test.js
marcelomorgado: Yes

packages/tasit-action/src/Contract.test.js
pcowgill: If you agree, feel free to resolve it

packages/tasit-action/src/Contract.test.js
pcowgill: Since you gave this a thumbs up, I'm resolving it.

packages/tasit-action/src/ContractEventsSubscription.js
pcowgill: If you agree it's Outdated and no longer relevant, feel free to resolve it

packages/tasit-action/src/TransactionSubscription.js
marcelomorgado: I've created separated issues https://github.com/tasitlabs/TasitSDK/issues/163 and https://github.com/tasitlabs/TasitSDK/issues/164

packages/tasit-action/src/Contract.test.js
marcelomorgado: Deprecated. The comment was deleted. I'm resolving this.

packages/tasit-action/src/Contract.test.js
marcelomorgado: I agree.

packages/tasit-action/src/ContractEventsSubscription.js
marcelomorgado: Deprecated. Contract object it now subscribing to the ContractEventsSubscription error event.

packages/tasit-action/src/ContractEventsSubscription.js
marcelomorgado: Resolving.

packages/tasit-action/src/ContractEventsSubscription.js
pcowgill: Great. Resolving.

packages/tasit-action/src/Contract.test.js
pcowgill:
```suggestion
```

packages/tasit-action/src/TransactionSubscription.js
pcowgill: Thanks. With that, should we resolve this?

packages/tasit-action/src/NFT.test.js
marcelomorgado: I've just created a function `confirmBalances(addresses, balances)`

packages/tasit-action/src/NFT.test.js
marcelomorgado: I've just refactored, is it better now?

packages/tasit-action/src/TransactionSubscription.js
marcelomorgado: Deprecated

packages/tasit-action/src/Contract.js
marcelomorgado: Great! Could we extract this thread to: https://github.com/tasitlabs/TasitSDK/issues/164 ?

packages/tasit-action/src/Contract.js
marcelomorgado: Done

packages/tasit-action/src/Subscription.js
marcelomorgado: Oh, It has no utility at all. Removed.

packages/tasit-action/src/Subscription.js
marcelomorgado: Done

packages/tasit-action/src/NFT.test.js
pcowgill: Great. Resolving this.

packages/tasit-action/src/TransactionSubscription.js
pcowgill: Good to know, thanks. Resolving this.

packages/tasit-action/src/Contract.js
pcowgill: Yep!

packages/tasit-action/src/Subscription.js
pcowgill: Great, thanks. Resolving this.

packages/tasit-action/src/Contract.js
pcowgill: Great, thanks. Resolving this.

packages/tasit-action/src/Subscription.js
pcowgill: Great, thanks. Resolving this.

packages/tasit-action/src/NFT.test.js
pcowgill: I'll take a look!

packages/tasit-action/src/Contract.test.js
pcowgill: Ah ok. I think I would prefer explicitly looking for both error and confirmation then, as opposed to just having it include confirmation.

packages/tasit-action/src/TransactionSubscription.js
marcelomorgado: I think so, resolving.

packages/tasit-action/src/Subscription.js
pcowgill: Hmm yeah. I'll have to think about this more, probably as part of this thread https://github.com/tasitlabs/TasitSDK/issues/164

Should we extract it there?

packages/tasit-action/src/NFT.test.js
pcowgill: Yeah, I think it's better now. Let's aim for a rule of a beforeEach being no more than 20 lines

packages/tasit-action/src/Contract.test.js
marcelomorgado: Oh, my fault. I've just replaced all `includes` to `deep.equal`.

packages/tasit-action/src/Subscription.js
marcelomorgado: Done.

packages/tasit-action/src/NFT.test.js
marcelomorgado: I agree, `Contract.test.js` stills having large hooks (> 20) but I think that will be better after https://github.com/tasitlabs/TasitSDK/issues/149

packages/tasit-action/src/Contract.test.js
pcowgill: Thanks. Resolving this.

packages/tasit-action/src/Subscription.js
pcowgill: Thanks. Resolving this

packages/tasit-action/src/NFT.test.js
pcowgill: Yep! Resolving this

pcowgill:
I’ll review the latest changes quickly and then merge this

Issue #180, opened 2019-01-31
Update Decentraland contracts


marcelomorgado:
Refs: https://decentraland.org/blog/technology/land-estates-registries-update

marcelomorgado:
I think that issue can be closed since I've already done that on PR https://github.com/tasitlabs/TasitSDK/pull/188

PR #188, opened 2019-02-04
PR 183 open threads and safeExecuteOrder

### LINKED ISSUE #186: Use safeExecuteOrder instead of executeOrder in Decentraland integration tests

marcelomorgado:
### Auto-close the issue?

Closes 
#186
#180 
https://github.com/tasitlabs/TasitSDK/pull/183#discussion_r252841129
https://github.com/tasitlabs/TasitSDK/pull/183#discussion_r253278692
https://github.com/tasitlabs/TasitSDK/pull/183#discussion_r253278715

### Types of changes

Docs (modified documentation or code comments)

## Note

I've needed to update Estate contract because older version didn't have `getFingerprint` used by `marketplace.safeTransferOrder`.


packages/tasit-sdk/src/TasitSdk.test.js
```diff
@@ -189,24 +209,28 @@ describe("Decentraland", () => {
...
         const allEstatesIds = [...Array(totalSupply.toNumber())].map(
           (val, key) => key + 1
         );
 
         for (let id of allEstatesIds) {
-          const order = await getEstateSellOrder(marketplace, estate, id);
+          const order = await getEstateSellOrder(
+            marketplaceContract,
+            estateContract,
+            id
```
pcowgill: Even id could be worth renaming to make it clear that it's the id of ____

packages/tasit-sdk/src/testHelpers/helpers.js
```diff
@@ -31,69 +31,84 @@ const gasParams = {
...
+const setupContracts = async ownerWallet => {
   // Note: It would be cooler to use NFT here if
   // Decentraland Land contract followed ERC721 exactly
-  const land = new Contract(landAddress, landABI, owner);
-  const landProxy = new Contract(landProxyAddress, landProxyABI, owner);
-  const estate = new Contract(estateAddress, estateABI, owner);
+  const landContract = new Contract(landAddress, landABI, ownerWallet);
+  const landProxyContract = new Contract(
+    landProxyAddress,
+    landProxyABI,
```
pcowgill: It's definitely the ABI of the land proxy and not the land contract?

pcowgill: I see elsewhere that we kept it as landABI

packages/tasit-sdk/src/testHelpers/helpers.js
```diff
@@ -31,69 +31,84 @@ const gasParams = {
...
-    marketplace,
+    manaContract,
+    landContract: landProxyContractWithLandABI,
+    estateContract,
+    marketplaceContract,
   };
 };
 
-const createParcels = async (landContract, lands, beneficiary) => {
+const createParcels = async (landContract, parcels, beneficiary) => {
```
pcowgill: Yeah, this is better since land is one of those nouns that is continuous and can't be made plural.

packages/tasit-sdk/src/testHelpers/helpers.js
marcelomorgado: There are both ABIs, after their initializations, the contract that is been exposing to test suite is the proxy one but with the land ABI.

packages/tasit-sdk/src/TasitSdk.test.js
marcelomorgado: Done

PR #193, opened 2019-02-05
Decentraland contracts testnet addresses

marcelomorgado:
### Issue link

Refs: #135

### Types of changes

### Notes

I'm not sure if this is enough to close the issue.

### DIFF
--- a/packages/tasit-contracts/decentraland/scripts/install.sh
+++ b/packages/tasit-contracts/decentraland/scripts/install.sh
@@ -9,26 +9,32 @@ fi
 
 DECENTRALAND_DIR="$PROJECT_DIR/decentraland";
 
-# Use list of repository_name:commit_hash
+# Use list of repository_name:commit_hash:is_npm
 REPOS="\
-    mana:f817fcab2c230bc2d2144ae85dff2bdc507ed659 \
-    land:f6795c4e47564b08d39d04f2f2d0f60fa97b69ac \
-    marketplace-contracts:190ab91887496da1677c41778832ff5a0c0c2ddb"
+    mana:f817fcab2c230bc2d2144ae85dff2bdc507ed659:true \
+    land:f6795c4e47564b08d39d04f2f2d0f60fa97b69ac:true \
+    marketplace-contracts:190ab91887496da1677c41778832ff5a0c0c2ddb:true \
+    contracts:d1ece59915692347a669bbf341ce6247a8a3406a:false";
 
 
 for clone in $REPOS;
 do
-    REPO=`echo $clone | awk -F ':' '{print $1}'`
-    COMMIT=`echo $clone | awk -F ':' '{print $2}'`
+    REPO=`echo $clone | awk -F ':' '{print $1}'`;
+    COMMIT=`echo $clone | awk -F ':' '{print $2}'`;
+    IS_NPM=`echo $clone | awk -F ':' '{print $3}'`;
+
     REPO_DIR="$DECENTRALAND_DIR/$REPO";
 
     if [ ! -e "$REPO_DIR/package.json" ];
     then
         rm -rf $REPO_DIR;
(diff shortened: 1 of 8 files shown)


packages/tasit-contracts/decentraland/scripts/install.sh
```diff
@@ -9,26 +9,32 @@ fi
...
     then
         rm -rf $REPO_DIR;
         git clone https://github.com/decentraland/$REPO.git $REPO_DIR;
-        cd $REPO_DIR && git fetch origin $COMMIT && git reset --hard FETCH_HEAD
+        cd $REPO_DIR && git fetch origin $COMMIT && git reset --hard FETCH_HEAD;
+    fi
 
+    if [ $IS_NPM ];
+    then
+        npm i --prefix $REPO_DIR;
```
pcowgill: What's the idea behind this `is_npm` in plain English? I think I see what's happening here, but I wanted to make sure.

packages/tasit-sdk/src/testHelpers/helpers.js
```diff
@@ -8,11 +8,15 @@ import { abi as landProxyABI } from "./abi/LANDProxy.json";
...
 import { abi as markplaceABI } from "./abi/Marketplace.json";
 
-const manaAddress = "0xb32939da0c44bf255b9810421a55be095f9bb3f4";
-const landAddress = "0x6191bc768c2339da9eab9e589fc8bf0b3ab80975";
-const landProxyAddress = "0x773f11ed472aa43e4ebaa963bcfbbea5a10c1bbd";
-const estateAddress = "0x41b598a2c618b59b74540ac3afffb32f7971b37a";
-const marketplaceAddress = "0x07c0e972064e5c05f7b3596d81de1afd35457eae";
+import { local as localAddresses } from "../../../tasit-contracts/decentraland/addresses";
+const {
+  MANAToken: manaAddress,
```
pcowgill: I wonder if `MANA_ADDRESS` would make more sense here

pcowgill:
```suggestion
  MANAToken: MANA_ADDRESS,
```

pcowgill: I'm resolving this since the suggestion in another thread already covers this idea

packages/tasit-contracts/decentraland/scripts/install.sh
marcelomorgado: If the repo is a npm project, run `npm install`. The `contracts` repo isn't a npm project, because of that I've added this condition.

pcowgill: Oh all the other ones are? I could have sworn I couldn’t find them the other day. But I guess I could have checked their package.json first

When you say contracts repo do you mean `marketplace-contracts`

marcelomorgado: No, not really.
/contract is the Decentraland Contract Addresses JSON (https://github.com/decentraland/contracts) it's only a data repo.

marcelomorgado: Yes, all others repos are npm / truffle projects.

packages/tasit-sdk/src/testHelpers/helpers.js
marcelomorgado: Done.

packages/tasit-contracts/decentraland/scripts/install.sh
pcowgill: Ah ok, thanks for the explanation. Cool! I don’t think we can count on that being the case for other projects, but we might as well take advantage of it for this one!

packages/tasit-sdk/src/testHelpers/helpers.js
pcowgill: Thanks. Resolving.

PR #194, opened 2019-02-06
Setting error as ethers.js loglevel

marcelomorgado:
### Issue link

Refs: https://github.com/tasitlabs/tasit/pull/79#issuecomment-460835222

### Types of changes

### DIFF
--- a/packages/tasit-action/src/Contract.js
+++ b/packages/tasit-action/src/Contract.js
@@ -7,6 +7,10 @@ import ProviderFactory from "./ProviderFactory";
 import Subscription from "./Subscription";
 import Action from "./Action";
 
+// Log levels: debug, default, info, warn, error, off
+// See more: https://github.com/ethers-io/ethers.js/blob/527de7ba5e1d31bd7c166a78d0fa62b58bf50a54/src.ts/errors.ts
+ethers.errors.setLogLevel("error");
+
 export class Contract extends Subscription {
   #provider;
   #ethersContract;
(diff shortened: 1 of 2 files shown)


packages/tasit-action/src/Contract.js
```diff
@@ -7,6 +7,10 @@ import ProviderFactory from "./ProviderFactory";
 import Subscription from "./Subscription";
 import Action from "./Action";
 
+// Log levels: debug, default, info, warn, error, off
+// See more: https://github.com/ethers-io/ethers.js/blob/527de7ba5e1d31bd7c166a78d0fa62b58bf50a54/src.ts/errors.ts
```
pcowgill: Are you sure you want to link to a specific git commit?

marcelomorgado: Hmm, maybe isn't a good idea.
Has nothing about loglevels on `ethers.js` docs. Only "documentation" that I've found was code and a issue: https://github.com/ethers-io/ethers.js/issues/379#issuecomment-454563534

pcowgill: Ah ok, interesting

marcelomorgado: So, should I keep that link? Or just change to a non-specific commit?
https://github.com/ethers-io/ethers.js/blob/master/src.ts/errors.ts

pcowgill: Up to you. I can see merits of both ways.

Issue #199, opened 2019-02-11
Action: Decentraland contracts basic support


marcelomorgado:
Implement basic Decentraland contracts objects to encapsulate ABIs.
Questions:
1) Should we have `TasitSDK.Action.Decentraland.Marketplace` or `TasitSDK.Action.DecentralandMarketplace`.

2) Should we use exactly contract names (`MANAToken`, `EstateRegistry`, etc) or "friendly"/short names like `Mana` and `Estate`?

pcowgill:
> * Should we have `TasitSDK.Action.Decentraland.Marketplace` or `TasitSDK.Action.DecentralandMarketplace`.

I would vote for `TasitSDK.Action.Marketplace.Decentraland` - that way it's getting progressively more specific. And then we could add `TasitSDK.Action.Marketplace.OpenSea` in the future.

> * Should we use exactly contract names (`MANAToken`, `EstateRegistry`, etc) or "friendly"/short names like `Mana` and `Estate`?

I'd vote for friendly/short names. What do you think?

marcelomorgado:
> I would vote for `TasitSDK.Action.Marketplace.Decentraland` - that way it's getting progressively more specific. And then we could add `TasitSDK.Action.Marketplace.OpenSea` in the future.

Ohh, nice! And about other contracts?
`TasitSDK.Action.ERC721.Estate`, `TasitSDK.Action.ERC721.Land` and `TasitSDK.Action.ERC20.Mana` 
or
`TasitSDK.Action.Decentraland.Estate`, `TasitSDK.Action.Decentraland.Land` and `TasitSDK.Action.Decentraland.Mana` 

> I'd vote for friendly/short names. What do you think?
Yes, I agree w/ you.

pcowgill:
I'm leaning towards this:

`TasitSDK.Action.ERC721.Decentraland.Estate`
`TasitSDK.Action.ERC721.Decentraland.Land`
`TasitSDK.Action.ERC20.Mana`

A project with only one ERC721 one type might skip the key for project name.

I almost went with NFT over ERC721, but I think it's nice to be consistent.

PR #200, opened 2019-02-11
Decentraland contracts

marcelomorgado:
### Auto-close the issue?

Closes https://github.com/tasitlabs/TasitSDK/issues/199

### Types of changes

### DIFF
--- a/packages/tasit-action/src/DecentralandEstate.test.js
+++ b/packages/tasit-action/src/DecentralandEstate.test.js
@@ -4,6 +4,8 @@ import { local as localAddresses } from "../../tasit-contracts/decentraland/addr
 const { EstateRegistry: ESTATE_ADDRESS } = localAddresses;
 
 describe("TasitAction.DecentralandEstate", () => {
+  // Note: This test runs against an unitialized contract.
+  // TODO: Improve that test case when we move to forked from testnet blockchain.
   it("should get the Estate owner", async function() {
     const estate = new DecentralandEstate(ESTATE_ADDRESS);
     let owner = await estate.owner();
--- a/packages/tasit-action/src/DecentralandEstate.test.js
+++ b/packages/tasit-action/src/DecentralandEstate.test.js
@@ -1,0 +1,12 @@
+import DecentralandEstate from "./DecentralandEstate";
+
+import { local as localAddresses } from "../../tasit-contracts/decentraland/addresses";
+const { EstateRegistry: ESTATE_ADDRESS } = localAddresses;
+
+describe("TasitAction.DecentralandEstate", () => {
+  it("should get the Estate owner", async function() {
+    const estate = new DecentralandEstate(ESTATE_ADDRESS);
+    let owner = await estate.owner();
+    expect(owner).to.equal("0x0000000000000000000000000000000000000000");
+  });
+});
--- a/packages/tasit-action/src/DecentralandMana.js
+++ b/packages/tasit-action/src/DecentralandMana.js
@@ -2,8 +2,6 @@ import Contract from "./Contract";
 import manaTokenABI from "./abi/MANAToken.json";
 
 export default class DecentralandMana extends Contract {
-  #contract;
-
   constructor(address, wallet) {
     const abi = manaTokenABI;
     super(address, abi, wallet);
--- a/packages/tasit-action/src/DecentralandMana.js
+++ b/packages/tasit-action/src/DecentralandMana.js
@@ -1,0 +1,11 @@
+import Contract from "./Contract";
+import manaTokenABI from "./abi/MANAToken.json";
+
+export default class DecentralandMana extends Contract {
+  #contract;
+
+  constructor(address, wallet) {
+    const abi = manaTokenABI;
+    super(address, abi, wallet);
+  }
+}
--- a/package-lock.json
+++ b/package-lock.json
@@ -57,11 +57,11 @@
       },
       "dependencies": {
         "@babel/generator": {
-          "version": "7.2.2",
-          "resolved": "https://registry.npmjs.org/@babel/generator/-/generator-7.2.2.tgz",
-          "integrity": "sha512-I4o675J/iS8k+P38dvJ3IBGqObLXyQLTxtrR4u9cSUJOURvafeEWb/pFMOTwtNrmq73mJzyF6ueTbO1BtN0Zeg==",
+          "version": "7.3.2",
+          "resolved": "https://registry.npmjs.org/@babel/generator/-/generator-7.3.2.tgz",
+          "integrity": "sha512-f3QCuPppXxtZOEm5GWPra/uYUjmNQlu9pbAD8D/9jze4pTY83rTtB1igTBSwvkeNlC5gR24zFFkz+2WHLFQhqQ==",
           "requires": {
-            "@babel/types": "^7.2.2",
+            "@babel/types": "^7.3.2",
             "jsesc": "^2.5.1",
             "lodash": "^4.17.10",
             "source-map": "^0.5.0",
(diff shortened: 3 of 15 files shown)


pcowgill:
@marcelomorgado Awesome! I'll take a look.

Quick initial reaction: I'm thinking `packages/tasit-action/src/abi/Marketplace.json` would fit more naturally in `packages/tasit-contracts/src/abi/`. We could still use it from within `tasit-action` or even re-expose the JSON within `tasit-action` in a DRY way.

packages/tasit-action/src/DecentralandEstate.test.js
```diff
@@ -0,0 +1,12 @@
+import DecentralandEstate from "./DecentralandEstate";
+
+import { local as localAddresses } from "../../tasit-contracts/decentraland/addresses";
```
pcowgill: Good candidates for a named package.json and importing using an `@name`

packages/tasit-action/src/DecentralandEstate.test.js
```diff
@@ -0,0 +1,12 @@
...
+const { EstateRegistry: ESTATE_ADDRESS } = localAddresses;
+
+describe("TasitAction.DecentralandEstate", () => {
+  it("should get the Estate owner", async function() {
+    const estate = new DecentralandEstate(ESTATE_ADDRESS);
+    let owner = await estate.owner();
+    expect(owner).to.equal("0x0000000000000000000000000000000000000000");
```
pcowgill: Why is this what ends up being the owner? Because it's unowned? Or is that the default with ganache?

pcowgill: And how does it relate to the land contract having an owner?

packages/tasit-action/src/DecentralandMana.js
```diff
@@ -0,0 +1,11 @@
+import Contract from "./Contract";
+import manaTokenABI from "./abi/MANAToken.json";
+
+export default class DecentralandMana extends Contract {
+  #contract;
+
+  constructor(address, wallet) {
```
pcowgill: I suppose we could consider DRYing this up with other classes where we do this, but I think that's probably DRY-overkill. So I'm leaning towards no change here.

packages/tasit-action/src/DecentralandEstate.test.js
marcelomorgado: It's unowned at this point because this test runs against an uninitialized contract.

packages/tasit-action/src/DecentralandEstate.test.js
marcelomorgado: OK

marcelomorgado:
> @marcelomorgado Awesome! I'll take a look.
> 
> Quick initial reaction: I'm thinking `packages/tasit-action/src/abi/Marketplace.json` would fit more naturally in `packages/tasit-contracts/src/abi/`. We could still use it from within `tasit-action` or even re-expose the JSON within `tasit-action` in a DRY way.

Yes, I agree. Test code can access the ABIs from `tasit-contract` using file system path, but the code the will be exposed don't. Should we have `tasit-contract` as `tasit-action` dependency?

packages/tasit-action/src/DecentralandMana.js
marcelomorgado: I've just realized that that `#contract` field is unnecessary.

pcowgill: Yeah, true!

packages/tasit-action/src/DecentralandEstate.test.js
pcowgill: Ah, right. Okay. Should we make a TODO to change that, or does the test serve its purpose fine with it uninitialized?

pcowgill:
> Yes, I agree. Test code can access the ABIs from `tasit-contract` using file system path, but the code the will be exposed don't. Should we have `tasit-contract` as `tasit-action` dependency?

Hmm, are you sure the code that will be exposed won't? I thought lerna handled this for us with some clever scripts.

I think having tasit-contract as a tasit-action dependency could make sense, yes. What do you think?

packages/tasit-action/src/DecentralandEstate.test.js
marcelomorgado: Seems that using package.json files to create "submodules" is a `react` feature.
Some refs:
https://stackoverflow.com/questions/49241941/react-native-absolute-import-with-alias-name-in-package-json
http://davidboyne.co.uk/2016/04/29/react-webpack-gem.html
https://medium.com/@sherryhsu/how-to-change-relative-paths-to-absolute-paths-for-imports-32ba6cce18a5
https://github.com/entwicklerstube/babel-plugin-root-import
https://www.npmjs.com/package/app-module-path
https://www.npmjs.com/package/babel-plugin-relative-path-import

I can keep looking but utils now, this kind of method that works with node/babel need module-directory mapping.

marcelomorgado:
> > Yes, I agree. Test code can access the ABIs from `tasit-contract` using file system path, but the code the will be exposed don't. Should we have `tasit-contract` as `tasit-action` dependency?
> 
> Hmm, are you sure the code that will be exposed won't have access to it? I thought lerna handled this for us with some clever scripts.

I'm not sure but I'm not think so and I don't found about that on `lerna` repo. We can publish and see if works.
>
> I think having `tasit-contract` as a `tasit-action` dependency could make sense, yes. What do you think?

If lerna doesn't resolving imports I think that is the way to go, It's seems a little odd having a `truffle` project as dependency but I can't find a better solution for now.

packages/tasit-action/src/DecentralandEstate.test.js
pcowgill: Ah right - okay, you can ignore this suggestion

pcowgill: If you agree, feel free to resolve this

pcowgill:
> I'm not sure but I'm not think so and I don't found about that on `lerna` repo. We can publish and see if works.

Sure, I'm up for running an experiment. We can always publish a newer version right after that (note that we wouldn't want to do this once this is used heavily by other projects)

> If lerna doesn't resolving imports I think that is the way to go, It's seems a little odd having a `truffle` project as dependency but I can't find a better solution for now.

Okay, sounds good

pcowgill:
@marcelomorgado Should we merge this now or make any of the changes mentioned above first?

packages/tasit-action/src/DecentralandMana.js
pcowgill: Resolving this since you addressed it in a newer commit

packages/tasit-action/src/DecentralandEstate.test.js
marcelomorgado: I think that serves for now, but I've added a comment to improve that when we move to a forked from testnet blockchain.

marcelomorgado:
> > I'm not sure but I'm not think so and I don't found about that on `lerna` repo. We can publish and see if works.
> 
> Sure, I'm up for running an experiment. We can always publish a newer version right after that (note that we wouldn't want to do this once this is used heavily by other projects)
> 
> > If lerna doesn't resolving imports I think that is the way to go, It's seems a little odd having a `truffle` project as dependency but I can't find a better solution for now.
> 
> Okay, sounds good

Ok, I've created an issue to track that: https://github.com/tasitlabs/TasitSDK/issues/201

marcelomorgado:
> @marcelomorgado Should we merge this now or make any of the changes mentioned above first?

I think that it's ready for merge now.

packages/tasit-action/src/DecentralandEstate.test.js
pcowgill: Okay, sounds good. Note that I still think we'll keep our original unit tests the same even once we've done that

packages/tasit-action/src/DecentralandEstate.test.js
```diff
@@ -0,0 +1,14 @@
+import DecentralandEstate from "./DecentralandEstate";
+
+import { local as localAddresses } from "../../tasit-contracts/decentraland/addresses";
+const { EstateRegistry: ESTATE_ADDRESS } = localAddresses;
+
+describe("TasitAction.DecentralandEstate", () => {
+  // Note: This test runs against an unitialized contract.
```
pcowgill:
```suggestion
  // Note: This test runs against an uninitialized contract.
```

PR #226, opened 2019-02-28
Action: Throw error event when tx.status = 0

### LINKED ISSUE #206: Action: Throw error event when tx.status = 0
The current code doesn't emit an error event when a transaction was broadcast successfully and then fails (e.g. failed call from a contract to another).

Olivierlu1:
### Issue link
https://github.com/tasitlabs/TasitSDK/issues/206

### Auto-close the issue?

Closes #206  

### Types of changes

In the Action.js file, the baseEthersListener doesn't emit an error when the transaction has failed (ie. status is 0). 

### Notes

I was not sure on what error message to emit, so I just wrote a standard one. Please feel free to change it to your liking.

### DIFF
--- a/packages/tasit-action/src/contract/Action.js
+++ b/packages/tasit-action/src/contract/Action.js
@@ -77,7 +77,10 @@ export class Action extends Subscription {
     const baseEthersListener = async blockNumber => {
       try {
         if (!this.#tx) this.#tx = await this.#txPromise;
-
+        
+        if (this.#tx.status == 0) {
+          throw new Error(`Transaction has failed and it's status is 0`);
+        }
         const receipt = await this.#provider.getTransactionReceipt(
           this.#tx.hash
         );


packages/tasit-action/src/contract/Action.js
```diff
@@ -77,7 +77,10 @@ export class Action extends Subscription {
     const baseEthersListener = async blockNumber => {
       try {
         if (!this.#tx) this.#tx = await this.#txPromise;
-
+        
+        if (this.#tx.status == 0) {
```
marcelomorgado: Hi @Olivierlu1 , we are trying to use [destructuring assignment](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Operators/Destructuring_assignment) whenever is possible, fell free to do that with `this.#tx` too.

packages/tasit-action/src/contract/Action.js
```diff
@@ -77,7 +77,10 @@ export class Action extends Subscription {
...
+          throw new Error(`Transaction has failed and it's status is 0`);
```
marcelomorgado: The error should be emitted to `errorListener` to be caught by the user instead of `throw` as an exception.
Use: `this._emitErrorEventFromEventListener()` for that.

pcowgill:
@Olivierlu1 Let's try to name the PRs based on what problem it solves rather than the number of the issue it solves. I'll rename this one. Thanks!

marcelomorgado:
Great work @Olivierlu1 , thanks! I've requested a few chages.

packages/tasit-action/src/contract/Action.js
marcelomorgado: Hey @Olivierlu1, I've just realized that destructuring syntax isn't works with https://github.com/tc39/proposal-class-fields private fields.
So, the syntax `const { #tx } = this;` won't work.

pcowgill: @marcelomorgado Good catch. I'd vote for moving to `const tx = this.#tx`

pcowgill:
@marcelomorgado Should we close this PR since the goal of these changes was achieved in https://github.com/tasitlabs/TasitSDK/pull/254?

marcelomorgado:
> @marcelomorgado Should we close this PR since the goal of these changes was achieved in #254?

Yes, I think so.

PR #240, opened 2019-03-01
Provider factory test suite improvement and bug fix

### LINKED ISSUE #208: TasitAction.ProviderFactory didn't work with networks different than local
Write more test cases and fix the current bug that it isn't working with ropsten network (The provider doesn't accept ropsten as network on config file).

marcelomorgado:
### Issue link

#208 

### Auto-close the issue?

Closes #208

### Types of changes

### DIFF
--- a/packages/tasit-action/src/ConfigLoader.test.js
+++ b/packages/tasit-action/src/ConfigLoader.test.js
@@ -84,6 +84,7 @@ const fulfillWithDefaults = config => {
     }
 
     provider = { ...provider, infura };
+
   } else if (providerType === "jsonrpc") {
     const { user, password, allowInsecure } = jsonRpc;
     jsonRpc = { ...jsonRpc, user, password };
--- a/packages/tasit-action/src/ConfigLoader.test.js
+++ b/packages/tasit-action/src/ConfigLoader.test.js
@@ -1,40 +1,98 @@
 import ProviderFactory from "./ProviderFactory";
 import ConfigLoader from "./ConfigLoader";
 
-// Note: This test suite needs to be improved
+const defaultConfig = ConfigLoader.getConfig();
+
+const extractConfigFromProvider = async ethersProvider => {
+  await ethersProvider.ready;
+  const network = await ethersProvider.getNetwork();
+  let { name: networkName } = network;
+  const {
+    pollingInterval,
+    connection,
+    apiToken,
+    apiAccessToken,
+    provider,
+    providers,
+    path,
+  } = ethersProvider;
+  const isJsonRpc = connection !== undefined;
+  const isEtherscan = apiToken !== undefined;
+  const isInfura = apiAccessToken !== undefined;
+  const isWeb3 = provider !== undefined;
+  const isFallback = providers !== undefined;
+  const isIpc = path !== undefined;
+
+  if (networkName === "unknown") networkName = "other";
+
+  let configProvider = {
+    network: networkName,
+    pollingInterval,
+  };
+
+  if (isJsonRpc) {
+    const { url, user, password, allowInsecure } = connection;
+    const { protocol, hostname, port } = new URL(url);
+    const jsonRpc = {
+      url: `${protocol}//${hostname}`,
+      port: Number(port),
+      user,
+      password,
+      allowInsecure,
+    };
+    configProvider = { ...configProvider, provider: "jsonrpc", jsonRpc };
+  }
+
+  const config = { provider: configProvider, events: { timeout: 2000 } };
+  return config;
+};
+
+const fulfillWithDefaults = config => {
+  let { provider } = config;
+  let { jsonRpc } = provider;
+  const isJsonRpc = jsonRpc !== undefined;
+  if (isJsonRpc) {
+    const { user, password, allowInsecure } = jsonRpc;
+    if (allowInsecure === undefined)
+      jsonRpc = { ...jsonRpc, user, password, allowInsecure: false };
+
+    provider = { ...provider, jsonRpc };
+  }
+
+  config = { ...config, provider };
+  return config;
+};
+
+const checkConfig = async config => {
+  ConfigLoader.setConfig(config);
+  const provider = ProviderFactory.getProvider();
+  const configFromProvider = await extractConfigFromProvider(provider);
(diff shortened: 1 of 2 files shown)


packages/tasit-action/src/ConfigLoader.test.js
```diff
@@ -1,40 +1,230 @@
...
+import { ethers } from "ethers";
+const { providers } = ethers;
+const {
+  JsonRpcProvider,
+  FallbackProvider,
+  InfuraProvider,
+  EtherscanProvider,
+} = providers;
+
+const parseNetworkNameFromEthres = networkName => {
```
pcowgill:
```suggestion
const parseNetworkNameFromEthers = networkName => {
```

packages/tasit-action/src/ConfigLoader.test.js
```diff
@@ -1,40 +1,230 @@
...
+    pollingInterval,
+    apiAccessToken: infuraApiKey,
+    apiKey: etherscanApiKey,
+    connection,
+    providers,
+    path,
+  } = provider;
+
+  let config = {
+    network: parseNetworkNameFromEthres(networkName),
```
pcowgill:
```suggestion
    network: parseNetworkNameFromEthers(networkName),
```

packages/tasit-action/src/ConfigLoader.test.js
```diff
@@ -1,40 +1,230 @@
...
+      url: `${protocol}//${hostname}`,
+      port: Number(port),
+      user,
+      password,
+      allowInsecure,
+    };
+    config = { ...config, provider: "jsonrpc", jsonRpc };
+  } else if (provider instanceof FallbackProvider) {
+    config = { ...config, provider: "fallback" };
+  }
```
pcowgill: Should we add an else case?

packages/tasit-action/src/ConfigLoader.test.js
```diff
@@ -1,40 +1,230 @@
...
+
+  if (providerType === "etherscan") {
+    if (!etherscan) etherscan = { apiKey: null };
+    else {
+      const { apiKey } = etherscan;
+      etherscan = { ...etherscan, apiKey };
+    }
+
+    provider = { ...provider, etherscan };
+  } else if (providerType === "infura") {
```
pcowgill:
```suggestion

  } else if (providerType === "infura") {
```

packages/tasit-action/src/ConfigLoader.test.js
```diff
@@ -1,40 +1,230 @@
...
+    if (!infura) infura = { apiKey: null };
+    else {
+      const { apiKey } = infura;
+      infura = { ...infura, apiKey };
+    }
+
+    provider = { ...provider, infura };
+  } else if (providerType === "jsonrpc") {
```
pcowgill:
```suggestion

  } else if (providerType === "jsonrpc") {
```

packages/tasit-action/src/ConfigLoader.test.js
```diff
@@ -1,40 +1,230 @@
...
+
+    provider = { ...provider, jsonRpc };
+  }
+
+  config = { ...config, provider };
+  return config;
+};
+
+const checkConfig = async config => {
+  const userConfig = fulfillWithDefaults(config);
```
pcowgill: Maybe a less vague word than fulfill? Or add what we're fulfilling to the variable name?

packages/tasit-action/src/ConfigLoader.test.js
```diff
@@ -1,40 +1,230 @@
...
+          network: "ropsten",
+          provider: "infura",
+          pollingInterval: 4000,
+        },
+        events: {
+          timeout: 2000,
+        },
+      };
+
+      await checkConfig(config);
```
pcowgill: Do we need to worry about this test suite running too often and failing intermittently because of that?

packages/tasit-action/src/ConfigLoader.test.js
marcelomorgado: Is `fulfillConfigWithDefaults` a better name?

packages/tasit-action/src/ConfigLoader.test.js
marcelomorgado: Sorry, I don't get it.

marcelomorgado:
Note: Some returns were undone by linter.

packages/tasit-action/src/ConfigLoader.test.js
pcowgill: Yeah, I like that better

packages/tasit-action/src/ConfigLoader.test.js
pcowgill: Does our unit test suite that runs in CI hit infura? It looks like it. If so, is there a risk of being throttled if we open too many PRs, for instance?

As a general rule keeping unit tests local seems like a goal worth striving for

pcowgill: But I can see why this would be a useful exception to that rule. Just food for thought, no action required before merging

pcowgill:
> Note: Some returns were undone by linter.

Ah, I wondered about that. Maybe prettier allows newlines after the `else if` line rather than before it?

packages/tasit-action/src/ConfigLoader.test.js
marcelomorgado: Hmmm I see. Yes, that test cases are hitting Infura and Etherscan. Maybe move both to integration tests in future?

marcelomorgado:
> > Note: Some returns were undone by linter.
> 
> Ah, I wondered about that. Maybe prettier allows newlines after the `else if` line rather than before it?

I've tested here and no new lines (after and before) are allowed.

packages/tasit-action/src/ConfigLoader.test.js
pcowgill: Yep, sounds good!

pcowgill:
> > > Note: Some returns were undone by linter.
> > 
> > 
> > Ah, I wondered about that. Maybe prettier allows newlines after the `else if` line rather than before it?
> 
> I've tested here and no new lines (after and before) are allowed.

Ah ok - good to know

PR #249, opened 2019-03-10
Decentraland deployment and tests against testnet

marcelomorgado:
### Issue link

#135 

### Types of changes

Meta (a change that affects future changes to this repo)

Technical debt (a code change that doesn't fix a bug or add a feature but makes something clearer for devs)

### Notes

Hey Paul,

Sorry for the delay, but I've stuck many times during these changes. I'll make an effort to delivery smallest / daily-basis PRs from now.
The last problem that I've faced is that: https://github.com/ethers-io/ethers.js/issues/453. After deployment of the contracts on Goerli testnet, the decentraland test suite failed against the ganache forked from testnet. I thought it might be because of the gorli testnet and I've deployed the contracts and run the test suite against Ropsten, but with the same problem. I kept both on our test cases because that can be useful if more issues occur. What do you think?

Summary of changes:

- Move contracts setups / data load from test suite to migrations;
- Using 3rd-parties truffle projects to run migrations (Until https://github.com/trufflesuite/truffle/issues/1799 stills open)
	Now, beyond migrations scripts, the truffle-config files are being overridden too.
- Improve migration (shell) script to work with testnets.
- Test suites were merged and improved to work with local / forked blockchains.
- Test suite was improved with missing test cases (buy land wasn't written before because ropsten hasn't a land parcel open sell order).

Note: Even using `git mv` in all moves and renames, I just noticed that some files changes weren't properly traced (I think that it could be because I've renamed and made changes on the same commit, sorry about that).

PR #251, opened 2019-03-12
GnosisSafe write functions to sync

### LINKED ISSUE #234: IdentityContract > GnosisSafe - change #executeTransaction to sync
GnosisSafe functions (transferERC20, transferNFT, transferEther and addSignerWithThreshold) should be changed to `sync` and keep returning `Action` object.

marcelomorgado:
### Issue link

#234 

### Auto-close the issue?

Closes #234 

### Types of changes

### DIFF
--- a/packages/tasit-identity-contract/src/GnosisSafe.js
+++ b/packages/tasit-identity-contract/src/GnosisSafe.js
@@ -1,5 +1,5 @@
-import Action from "tasit-action";
-const { Contract, ERC20, ERC721 } = Action;
+import TasitAction from "tasit-action";
+const { Action, Contract, ERC20, ERC721 } = TasitAction;
 const { ERC20Detailed } = ERC20;
 const { ERC721Full } = ERC721;
 import GnosisSafeUtils from "./GnosisSafeUtils";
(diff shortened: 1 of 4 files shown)


packages/tasit-identity-contract/src/GnosisSafe.js
```diff
@@ -61,59 +61,55 @@ export default class GnosisSafe extends Contract {
...
     const etherValue = "0";
-    const action = await this.#executeTransaction(data, to, etherValue);
+    const action = this.#executeTransaction(data, to, etherValue);
+    return action;
+  };
+
+  #executeTransaction = (data, toAddress, etherValue) => {
+    const tx = this.#prepareTransaction(data, toAddress, etherValue);
+    const provider = this._getProvider();
+    const action = new Action(tx, provider);
```
pcowgill: Will an action always need a provider? Maybe not if it's a meta-tx.

marcelomorgado: For now - since all tx are broadcasted to the blockchain - yes. The Action class should be refactored when we start broadcasting to off chain relayers.

pcowgill: Okay, that's what I thought. I'll resolve this and we can just keep that in mind for when it comes up again. Thanks.

pcowgill:
@marcelomorgado Any strong preference whether we merge this before or after https://github.com/tasitlabs/TasitSDK/pull/249?

marcelomorgado:
> @marcelomorgado Any strong preference whether we merge this before or after #249?

I think that can be merged now since none of these four files are been touched by PR #249

PR #254, opened 2019-03-12
GnosisSafe insufficient signers testcase

### LINKED ISSUE #206: Action: Throw error event when tx.status = 0

### LINKED ISSUE #232: IdentityContract > GnosisSafe: Create test case for transaction with insufficient signers
Current Gnosis Safe test suite is testing a wallet with one signer.
Change that to have more than one signer and create a test case that sends a transaction with less signatures than the threshold.

Note: `tasit-identity-contract` will be renamed.

Refs: https://github.com/tasitlabs/TasitSDK/pull/217#discussion_r260878190

### LINKED ISSUE #233: TasitAction: Create a test case to test failed transaction
The issue https://github.com/tasitlabs/TasitSDK/issues/206 will implement error event emission when a transaction sent by an action have status = 0.
A test case should be created after that issue is closed.

marcelomorgado:
### Issue link

#232 #206 #233 

### Auto-close the issue?

Closes #232 
Closes #206 
Closes #233 

### Types of changes

Other (please describe below if you select "Other")

### Notes

This test case demand solves these other issues. With this PR, the later PR #226 is now deprecated.

### DIFF
--- a/packages/tasit-identity-contract/src/GnosisSafe.test.js
+++ b/packages/tasit-identity-contract/src/GnosisSafe.test.js
@@ -155,7 +155,7 @@ describe("GnosisSafe", () => {
       expect(`${thresholdAfter}`).to.equal(`2`);
     });
 
-    it("shouldn't can execute transfer with insufficient signers", async () => {
+    it("shouldn't be able to execute transfer with insufficient signers", async () => {
       const onError = sinon.fake();
 
       await etherFaucet(provider, root, GNOSIS_SAFE_ADDRESS, ONE);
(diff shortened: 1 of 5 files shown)


packages/tasit-identity-contract/src/GnosisSafe.test.js
```diff
@@ -128,30 +128,66 @@ describe("GnosisSafe", () => {
...
+      const ownersAfter = await gnosisSafe.getOwners();
+      expect(ownersAfter).deep.equal([newSignerAddress, johnAddress]);
+
+      const thresholdAfter = await gnosisSafe.getThreshold();
+      expect(`${thresholdAfter}`).to.equal(`2`);
+    });
 
-    const ownersAfter = await gnosisSafe.getOwners();
-    expect(ownersAfter).deep.equal([newSignerAddress, johnAddress]);
+    it("shouldn't can execute transfer with insufficient signers", async () => {
```
pcowgill:
```suggestion
    it("shouldn't be able to execute transfer with insufficient signers", async () => {
```

PR #268, opened 2019-03-14
Truffle external compilers

### LINKED ISSUE #247: Use truffle external compiler setup
Truffle has a configuration option to use compilers from external projects that allow tracking, testing and write migration scripts from the main truffle project. That feature will help us because we can have 3rd-parties migrations scripts inside of `migrations` folder of `tasit-contracts` instead of override 3rd-parties scripts.
I've tried to use that feature with our 3rd-parties contracts, but I've faced some issues and seems that's buggy.
I've opened an issue explaining these in details: https://github.com/trufflesuite/truffle/issues/1799

marcelomorgado:
### Issue link

#247

### Auto-close the issue?

Closes #247 

### Types of changes

### Notes

I've spent almost a day on that but I think that will worth because the contracts interactions became harder to debug, from now we can use the `truffle debug` feature.


packages/tasit-contracts/.secret.json
```diff
@@ -0,0 +1,10 @@
+{
```
pcowgill: @marcelomorgado Did you mean for this to be `.gitignore`d?

packages/tasit-contracts/migrations/4_gnosis_contracts.js
```diff
@@ -0,0 +1,20 @@
...
+// accounts[9]
+const johnAddress = "0x8226bcef50b3c76a9eb7eba0c09ebbb2362e5db7";
+
+module.exports = function(deployer) {
+  // Workaround to write async/await migration scripts
+  // See more: https://github.com/trufflesuite/truffle/issues/501#issuecomment-332589663
+  deployer.then(async () => {
+    const proxyFactory = await deployer.deploy(ProxyFactory);
+    const safe = await deployer.deploy(GnosisSafe);
+    const owners = [johnAddress];
```
pcowgill: Should we still call them `john`?

pcowgill:
@marcelomorgado Great work! I agree that this will be very helpful.

packages/tasit-contracts/.secret.json
marcelomorgado: I've added this file to the `.gitignore`, I think that is being shown because I added after the commit of the `.secret.json`

marcelomorgado:
> @marcelomorgado Great work! I agree that this will be very helpful.

Thanks!

packages/tasit-contracts/migrations/4_gnosis_contracts.js
marcelomorgado: Oh, I'm not replying same name for all tests, should I? I'm trying to keep the wallet's names most clear as possible for each file/test suite.
Will be better if I called it as `signer` ?

packages/tasit-contracts/.secret.json
pcowgill: Ah ok - you may have to manually remove it then

packages/tasit-contracts/migrations/4_gnosis_contracts.js
pcowgill: What do you mean by "replying same name"?

pcowgill: Keeping the wallet's name as clear as possible for each file / test suite sounds good.

Sure, `signer` sounds like a good name

marcelomorgado: Using the same name to accounts among test suites.

marcelomorgado: Done.

packages/tasit-contracts/.secret.json
marcelomorgado: Done.

packages/tasit-contracts/migrations/4_gnosis_contracts.js
pcowgill: Ah ok - yeah I think the accounts can be used for different roles between different files / test suites

pcowgill: Resolving this

packages/tasit-contracts/.secret.json
pcowgill: Thanks! Resolving

PR #289, opened 2019-03-20
Start script for use with tasit repo

pcowgill:
### Issue link

n/a

### Types of changes

### Notes

I thought this would be useful when using the Tasit SDK primarily for development of apps in the `tasit` repo. Open to suggestions that this should be called something other than `start`


pcowgill:
Note: This is mostly useful after having pulled the latest changes from the `develop` branch of the Tasit SDK on GitHub and you want to make sure all the deps are installed before doing `npm run prepare:blockchain`

PR #292, opened 2019-03-22
Decentraland contracts data

### LINKED ISSUE #285: Prepare decentraland contracts data
- Prices should follow that rule: [a rand number between 10 to 100] * 1000 MANA.
- Avoid creation of assets without images (https://github.com/tasitlabs/tasit/issues/134).

Extracted from: https://github.com/tasitlabs/tasit/issues/156

### LINKED ISSUE #288: TasitSDK: Improve DecentralandUtils API
1. Move `getOpenSellOrders(fromBlock)` to `getAllAssetsForSale()`
2. For now, each item of the list is an event log item and the open order data is extracted from the `values` parameter. Make the function returns the open sell order data.

marcelomorgado:
### Issue link

#285 
#288 

### Auto-close the issue?

Closes #285
Closes #288 

### Types of changes


packages/tasit-sdk/package.json
```diff
@@ -46,6 +46,7 @@
     "@babel/register": "^7.0.0",
     "chai": "^4.2.0",
     "ethers": "^4.0.27",
-    "mocha": "^6.0.1"
+    "mocha": "^6.0.1",
+    "node-fetch": "^2.3.0"
```
pcowgill: Hmm, I think we only need it as a dep in one place, and then we can use it via `tasit-action` here?

packages/tasit-sdk/src/scripts/checkDecentralandContractsData.js
```diff
@@ -0,0 +1,83 @@
+//
+// Script used to check if the target chain has the Decentraland app pre-conditions
+//
+// Note: This script is using mocha for convenince but isn't a test suite to be runned by `test` script.
```
pcowgill:
```suggestion
// Note: This script is using mocha for convenience but isn't a test suite to be run by the `test` script.
```

packages/tasit-sdk/src/scripts/checkDecentralandContractsData.js
```diff
@@ -0,0 +1,83 @@
...
+// Use: npx mocha src/scripts/checkDecentralandContractsData.js  --require @babel/register --file src/testHelpers/mochaSetup.js --no-timeouts
```
pcowgill: Worth exposing this with an npm script?

packages/tasit-sdk/src/scripts/checkDecentralandContractsData.js
```diff
@@ -0,0 +1,83 @@
...
+
+import { Action, ContractBasedAccount } from "../TasitSdk";
+const { ERC20 } = Action;
+const { Mana } = ERC20;
+
+import DecentralandUtils from "../helpers/DecentralandUtils";
+
+const fetch = require("node-fetch");
+
+const { TWO, TEN, BILLION, WeiPerEther } = constants;
```
pcowgill: It feels weird that `WeiPerEther` has a different capitalization convention

packages/tasit-sdk/src/scripts/checkDecentralandContractsData.js
```diff
@@ -0,0 +1,83 @@
...
+
+describe("Decentraland App pre-conditions", () => {
+  const mana = new Mana(MANA_ADDRESS);
+  const gnosisSafeOwner = accounts[9];
```
pcowgill: Maybe some comment in the code to remind readers why it is `accounts[9]` that is the owner?

packages/tasit-sdk/src/scripts/checkDecentralandContractsData.js
```diff
@@ -0,0 +1,83 @@
...
+    });
+
+    it("should have at least a billion MANA", async () => {
+      await expectMinimumTokenBalances(mana, [GNOSIS_SAFE_ADDRESS], [BILLION]);
+    });
+  });
+
+  describe("Gnosis Safe wallet owner", () => {
+    it("should have at least ten ethers", async () => {
+      await expectMinimumEtherBalances(provider, [GNOSIS_SAFE_ADDRESS], [TWO]);
```
pcowgill: The test name is wrong if we're going to check for two rather than ten

packages/tasit-sdk/src/scripts/checkDecentralandContractsData.js
```diff
@@ -0,0 +1,83 @@
...
+  });
+
+  describe("Marketplace", () => {
+    it("should have at least 100 assets for sale", async () => {
+      expect(assetsForSale.length).to.be.at.least(100);
+    });
+  });
+
+  it("Assets for sale", async () => {
+    const minPrice = bigNumberify("10000").mul(WeiPerEther);
```
pcowgill: How will this work with preexisting assets from the last time this was run?

packages/tasit-sdk/src/scripts/checkDecentralandContractsData.js
```diff
@@ -0,0 +1,83 @@
...
+    const maxPrice = bigNumberify("100000").mul(WeiPerEther);
```
pcowgill: Is it bad to use `WeiPerEther` here since this is really more like the number of decimals of the token?

packages/tasit-sdk/src/scripts/checkDecentralandContractsData.js
```diff
@@ -0,0 +1,83 @@
...
+      // Parcels always show correct image
+      if (isEstate) {
+        const image = await fetch(
+          `https://api.decentraland.org/v1/estates/${assetId}/map.png`
+        );
+        const imageData = (await image.buffer()).toString("base64");
+        expect(
+          imageData,
+          `The image of the estate (id: ${assetId}) is blank`
+        ).not.equals(blankImageData);
```
pcowgill: Would this be a faster op if we did it on just part of the image data? Or is it fast enough as-is?

packages/tasit-sdk/src/scripts/checkDecentralandContractsData.js
```diff
@@ -0,0 +1,83 @@
...
+      const price = bigNumberify(priceInWei);
+
+      const isLand = addressesAreEqual(nftAddress, LAND_PROXY_ADDRESS);
+      const isEstate = addressesAreEqual(nftAddress, ESTATE_ADDRESS);
+
+      expect(price.gte(minPrice), `${price} isn't >= ${minPrice}`).to.be.true;
+      expect(price.lte(maxPrice), `${price} isn't <= ${maxPrice}`).to.be.true;
+
+      // Parcels always show correct image
+      if (isEstate) {
```
pcowgill: The comment isn't quite right if it's an image for an estate rather than a land parcel

packages/tasit-contracts/src/scripts/populateDecentralandContracts.js
```diff
@@ -23,35 +23,170 @@ import fs from "fs";
...
 
-const createMultipleParcels = async (
-  landContract,
-  parcels,
-  beneficiaryAddress,
-  contractOwnerWallet
-) => {
-  let xArray = [];
-  let yArray = [];
+let network = process.env.NETWORK;
```
pcowgill: Worth documenting somewhere that someone would need to set this? Or logging if it's not set?

packages/tasit-contracts/src/scripts/populateDecentralandContracts.js
```diff
@@ -23,35 +23,170 @@ import fs from "fs";
...
+
+const provider = ProviderFactory.getProvider();
+
+const cancelSellOrder = async (nftAddress, id) => {
+  const action = marketplaceContract.cancelOrder(
+    nftAddress,
+    `${id}`,
+    gasParams
+  );
+  await action.waitForNonceToUpdate();
```
pcowgill: Does a try/catch belong in here our outside where the `cancelSellOrder` function is used? I guess it depends if we want to return `false` or something if it doesn't work.

packages/tasit-contracts/src/scripts/populateDecentralandContracts.js
```diff
@@ -23,35 +23,170 @@ import fs from "fs";
...
+  // Note: Estate with 5 is one of the estates with blank image
+  const blankImageData = await getImageDataFromEstateId(5);
+
+  marketplaceContract.setWallet(sellerWallet);
+  for (let id of estatesIds) {
+    const imageData = await getImageDataFromEstateId(id);
+
+    if (imageData === blankImageData) {
+      console.log(
+        `Removing order of estate (id: ${id}) because it's with a blank image.`
```
pcowgill:
```suggestion
        `Removing order of estate (id: ${id}) because it has a blank image.`
```

packages/tasit-contracts/src/scripts/populateDecentralandContracts.js
```diff
@@ -23,35 +23,170 @@ import fs from "fs";
...
+    );
+    const data = (await image.buffer()).toString("base64");
+    return data;
+  };
+
+  // Note: Estate with 5 is one of the estates with blank image
+  const blankImageData = await getImageDataFromEstateId(5);
+
+  marketplaceContract.setWallet(sellerWallet);
+  for (let id of estatesIds) {
```
pcowgill: Worth considering doing this in parallel?

packages/tasit-contracts/src/scripts/populateDecentralandContracts.js
```diff
@@ -23,35 +23,170 @@ import fs from "fs";
...
+
+// Tech-debt: Use `assignMultipleParcels` to save gas cost.
+// The amount of parcels per call should be short enough to avoid out-of-gas.
+const createParcels = async allParcels => {
+  console.log("Creating parcels...");
+
+  let parcelIds = [];
+  for (let parcel of allParcels) {
+    try {
+      const id = await createParcel(parcel);
```
pcowgill: We could also move the metadata step here

packages/tasit-contracts/src/scripts/populateDecentralandContracts.js
```diff
@@ -23,35 +23,170 @@ import fs from "fs";
...
+    const { x, y } = parcel;
+    const id = await landContract.encodeTokenId(`${x}`, `${y}`);
+    parcelIds.push(id);
+  }
+
+  return parcelIds;
+};
+
+// Tech-debt: Use `assignMultipleParcels` to save gas cost.
+// The amount of parcels per call should be short enough to avoid out-of-gas.
```
pcowgill: Good point

packages/tasit-contracts/src/scripts/populateDecentralandContracts.js
```diff
@@ -23,35 +23,170 @@ import fs from "fs";
...
+  const action = landContract.assignNewParcel(
+    `${x}`,
+    `${y}`,
+    sellerAddress,
     gasParams
   );
-  await assignAction.waitForNonceToUpdate();
+
+  const parcelId = await new Promise((resolve, reject) => {
+    action.once("confirmation", async message => {
```
pcowgill: Why use this style here vs. `waitForNonceToUpdate` elsewhere?

packages/tasit-contracts/src/scripts/populateDecentralandContracts.js
```diff
@@ -78,263 +213,184 @@ const createEstate = async (estateContract, landContract, estate, wallet) => {
...
 
   await action.waitForNonceToUpdate();
-  console.log("created id =", estateId);
+  console.log(`created id ${estateId}`);
 
   return estateId;
 };
 
-const createEstates = async (estateContract, landContract, estates, wallet) => {
+const createEstates = async estates => {
```
pcowgill: So we're moving away from a purely functional style where all the args needed are passed in? I suppose that's fine for the purpose of this script, just noting it.

We could make a convenience function with just the one parameter call another function behind the scenes that grabs the other args from global constants so that each function remains more testable

packages/tasit-contracts/src/scripts/populateDecentralandContracts.js
```diff
@@ -78,263 +213,184 @@ const createEstate = async (estateContract, landContract, estate, wallet) => {
...
-      estateAddress,
-      assetId,
-      priceInWei,
-      expireAt,
-      gasParams
-    );
-    await action.waitForNonceToUpdate();
+function getRandomInt(min, max) {
+  min = Math.ceil(min);
+  max = Math.floor(max);
```
pcowgill: Why is this step necessary?

packages/tasit-contracts/src/scripts/populateDecentralandContracts.js
```diff
@@ -78,263 +213,184 @@ const createEstate = async (estateContract, landContract, estate, wallet) => {
...
+  return Math.floor(Math.random() * (max - min)) + min;
+}
+
+const placeAssesOrders = async (estateIds, parcelIds) => {
```
pcowgill:
```suggestion
const placeAssetOrders = async (estateIds, parcelIds) => {
```

packages/tasit-contracts/src/scripts/populateDecentralandContracts.js
```diff
@@ -78,263 +213,184 @@ const createEstate = async (estateContract, landContract, estate, wallet) => {
...
+  const { parcels: estateParcels } = estate;
+
+  for (let parcel of estateParcels)
+    if (findParcel(parcel, listOfParcels)) return true;
+
+  return false;
+};
+
+const getEstatesFromAPI = async () => {
+  const res = await fetch(
```
pcowgill: Add a comment in the code about how we'll want to move away from this point of centralization in the future

packages/tasit-contracts/src/scripts/populateDecentralandContracts.js
```diff
@@ -78,263 +213,184 @@ const createEstate = async (estateContract, landContract, estate, wallet) => {
...
+  const parcels = parcelsFromAPI.map(parcel => {
+    const { x, y, data } = parcel;
+    let { name: metadata } = data;
+    if (!metadata) metadata = "";
+    return { x, y, metadata };
+  });
 
-  const allParcels = [
-    // Estate: all road adjacent parcels
-    { x: -30, y: -105, metadata: `` },
```
pcowgill: If we make a unit test suite for this script, this data could be used to mock an API response

packages/tasit-contracts/src/scripts/populateDecentralandContracts.js
```diff
@@ -78,263 +213,184 @@ const createEstate = async (estateContract, landContract, estate, wallet) => {
...
+  const { data: jsonData } = json;
+  const { estates: estatesFromAPI } = jsonData;
+
+  const allEstates = estatesFromAPI.map(estate => {
+    const { data } = estate;
+    const { name: metadata, parcels } = data;
+    return { metadata, parcels };
+  });
+
+  // Keep estates with small number of parcels to avoid out-of-gas on creation
```
pcowgill: Good call

packages/tasit-contracts/src/scripts/populateDecentralandContracts.js
```diff
@@ -78,263 +213,184 @@ const createEstate = async (estateContract, landContract, estate, wallet) => {
...
 
-    // Update parcels with metadata
-    console.log("Updating parcels with metadata...");
-    landContract.setWallet(sellerWallet);
-    for (let parcel of allParcels) {
-      const { x, y, metadata: parcelName } = parcel;
-      const updateAction = landContract.updateLandData(x, y, parcelName);
-      await updateAction.waitForNonceToUpdate();
-    }
+    const estatesParcels = extractParcelsFromEstates(estatesFromAPI);
```
pcowgill: Why is there any need to extract these parcels instead of just dealing with the estates directly?

pcowgill: Is it because sometimes individual parcels from within the estate are also for sale? I don't know if that's necessarily a problem

packages/tasit-contracts/src/scripts/populateDecentralandContracts.js
```diff
@@ -78,263 +213,184 @@ const createEstate = async (estateContract, landContract, estate, wallet) => {
...
-    );
+    const parcelIds = await createParcels(parcelsToCreate);
+
+    await updateParcelsData(parcelsToCreate);
+
+    const estateIds = await createEstates(estatesToCreate);
+
+    await approveMarketplace();
+
+    await placeAssesOrders(estateIds, parcelIds);
```
pcowgill:
```suggestion
    await placeAssetOrders(estateIds, parcelIds);
```

pcowgill:
@marcelomorgado Great work! Note that very few of these comments if any need to gate merging. Let me know when you think it's ready to merge.

packages/tasit-action/package.json
```diff
@@ -38,6 +38,7 @@
     "chai": "^4.2.0",
     "chai-as-promised": "^7.1.1",
     "mocha": "^6.0.2",
+    "node-fetch": "^2.3.0",
```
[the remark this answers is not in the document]
marcelomorgado: Since I've added that as a `devDependency` only I wasn't worried about the better option for `react-native` app, I've just picked the one that seems most popular.
I just took a quick look at `fribee` and `axios` and I think that:
`frisbee` is a good choice for simple rest API consumption;
`axios` seems more complete and better for more complex needs.
If will be better to us pick one module that will be used on our apps too, I think that `axios` can be a better option since our needs aren't totally clear at the moment.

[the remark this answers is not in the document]
pcowgill: Thanks for the detailed thoughts. That all makes sense. If you like we can extract this to an issue and keep it as-is for now.

packages/tasit-contracts/src/scripts/populateDecentralandContracts.js
pcowgill: I should have made the comment one line lower so you could see that I mean the Decentraland API

packages/tasit-sdk/package.json
marcelomorgado: Note: We are using `node-fetch` as devDependency on `tasit-sdk` to test if an estates' image is a blank one.
Said that, I think that we really can add the dependency to one of the packages and it'll work for the others because of the `lerna --hoist` option:

> Lerna will hoist dependencies which are used in multiple projects, even if they are not used in all projects. [(refs)](https://github.com/lerna/lerna/blob/master/doc/hoist.md#forgetting-dependencies)

It's a trick that will work for `devDependencies` but it's putted as a `Disadvantages with hoisting` because it's a error prone if used with non-dev dependencies.

I'm not sure if I undestood well what you mean.

packages/tasit-sdk/src/scripts/checkDecentralandContractsData.js
marcelomorgado: Yes, for sure. But about run it within the test suite script? I'm not sure.

packages/tasit-sdk/src/scripts/checkDecentralandContractsData.js
marcelomorgado: Since we have a few open sell orders with 1MANA price and they will be executed soon, I'm assuming the new ones for dev env.

packages/tasit-sdk/src/scripts/checkDecentralandContractsData.js
marcelomorgado: Yeah, I agree.
`bigNumberify("10000").mul(1e18);` Seems better?

packages/tasit-contracts/src/scripts/populateDecentralandContracts.js
marcelomorgado: I've used that style only for `estate` and `parcel` creation because:
1. We need the asset id for sell order creation.
2. The script needs to keep going if an asset already exists.

packages/tasit-contracts/src/scripts/populateDecentralandContracts.js
marcelomorgado: Yes, I moved out from functional style just because it's a self-contained script and for this case seemed to me that the code will be clearer if functions have few args.
I like your idea to wrap functional functions :+1: I'll keep that in mind for the future.

packages/tasit-contracts/src/scripts/populateDecentralandContracts.js
marcelomorgado: `getRandonInt`? I'm using that for assets prices.

packages/tasit-contracts/src/scripts/populateDecentralandContracts.js
marcelomorgado: Yes, I agree. Worth keep that? It's no big deal since I've got these data in a few minutes looking for the Decentraland marketplace.

packages/tasit-contracts/src/scripts/populateDecentralandContracts.js
marcelomorgado: A) The parcels creation function receives a parcels list and I'm extracting those to call that function once. On the other hand, I can call several times (one time for each estate).

B) I'm removing duplicated parcels from an estate that already was getting from parcels API. On the other hand, the parcel creation function is already ignoring failed creation.

packages/tasit-action/package.json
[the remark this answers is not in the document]
marcelomorgado: Extracted to: https://github.com/tasitlabs/TasitSDK/issues/294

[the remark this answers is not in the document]
pcowgill: Thanks. Resolving this.

packages/tasit-sdk/src/scripts/checkDecentralandContractsData.js
marcelomorgado: I think that the http request is the slowest operation of each iteration. Should I make an analysis on that?

packages/tasit-sdk/src/scripts/checkDecentralandContractsData.js
marcelomorgado: I'll improve the comment.

packages/tasit-sdk/src/scripts/checkDecentralandContractsData.js
pcowgill: I agree, we wouldn't want it to run along with the test suite script

packages/tasit-contracts/src/scripts/populateDecentralandContracts.js
marcelomorgado: I'm using higher-level try/catch on that and other operation that a specific return isn't needed and event listener style on otherwise.

packages/tasit-sdk/src/scripts/checkDecentralandContractsData.js
pcowgill: Okay, that's fair. Let's make sure to execute them soon. Resolving this.

packages/tasit-sdk/src/scripts/checkDecentralandContractsData.js
pcowgill: @marcelomorgado Yeah, or maybe a new constant called `TOKEN_SUBDIVISIONS` that is equal to 1e18

packages/tasit-sdk/src/scripts/checkDecentralandContractsData.js
pcowgill: Probably overkill for now until something motivates the research. Feel free to resolve this.

packages/tasit-sdk/src/scripts/checkDecentralandContractsData.js
pcowgill: Thanks. Feel free to resolve this when that's done.

packages/tasit-contracts/src/scripts/populateDecentralandContracts.js
pcowgill: Got it - thanks. Resolving this.

packages/tasit-contracts/src/scripts/populateDecentralandContracts.js
pcowgill: Not gating for this PR, though.

Most likely we'd want to use a queue data structure where we only process up to n in parallel because we would get throttled if we did too much in parallel

packages/tasit-contracts/src/scripts/populateDecentralandContracts.js
pcowgill: Resolving this

packages/tasit-contracts/src/scripts/populateDecentralandContracts.js
pcowgill: Okay, that makes sense, thanks. Resolving this.

packages/tasit-sdk/src/scripts/checkDecentralandContractsData.js
marcelomorgado: Resolving this.

packages/tasit-contracts/src/scripts/populateDecentralandContracts.js
pcowgill: No, the `Math.floor`

packages/tasit-contracts/src/scripts/populateDecentralandContracts.js
pcowgill: Nah, we don't need to keep it. Good to know it's easy to get back (which makes sense). I'll resolve this.

packages/tasit-contracts/src/scripts/populateDecentralandContracts.js
pcowgill: Resolving this

packages/tasit-sdk/package.json
pcowgill: I was saying it wouldn't need to work for the others if we stuck to always using it indirectly via some function we expose from the original package it's used in

pcowgill: If we did want to use it directly in multiple spots, I think we could only count on hoisting without making it an explicit dependency everywhere we wanted it if it were used in at least 2 of n packages, not 1 of n

pcowgill: But I think exposing it via a helper function to use in the Tasit SDK tests or scripts might be too much work to DRY it up for now

packages/tasit-sdk/src/scripts/checkDecentralandContractsData.js
pcowgill: So we'll still expose this script as a non-test npm script, though?

packages/tasit-contracts/src/scripts/populateDecentralandContracts.js
pcowgill: Cool, sounds good! Resolving this

packages/tasit-contracts/src/scripts/populateDecentralandContracts.js
pcowgill: A) Ah yes, I forgot we would need to create the underlying parcels that belong to the estates in the test env. Makes a lot of sense now, though.

B) Ah ok, so it's just to avoid creating them twice. It's still nice from a code understandability perspective not to count on failed creation to keep them from being created. So I like the way you've done this.

C) We still list the parcels and estates for sale separately if a parcel is for sale that's also in an estate that's for sale, right? This was the topic I was originally thinking about when creating this thread.

This all makes sense to me now, and I think we can resolve it.

packages/tasit-sdk/package.json
marcelomorgado: Done.

packages/tasit-sdk/src/scripts/checkDecentralandContractsData.js
marcelomorgado: Yes, done.

packages/tasit-sdk/src/scripts/checkDecentralandContractsData.js
marcelomorgado: Done.

packages/tasit-sdk/package.json
pcowgill: Wait, so after the removal of the dep from the Tasit SDK package, aren't we in a 1-of-n situation now? So we'd need to change something else too, right?

packages/tasit-sdk/src/scripts/checkDecentralandContractsData.js
pcowgill: Thanks! Resolving this.

packages/tasit-sdk/src/scripts/checkDecentralandContractsData.js
pcowgill: Thanks! Resolving this.

packages/tasit-contracts/src/scripts/populateDecentralandContracts.js
marcelomorgado: A and B) :+1: 
C) I'm almost sure that if an estate is for sale, none of its parcels could have a sell order too.

Ok, resolving this.

packages/tasit-sdk/package.json
marcelomorgado: Hmmm since `lerna bootstrap` is made with the `--hoist` option, all deps are installed on `TasitSDK/node_modules` and accessible for all packages.

packages/tasit-sdk/src/scripts/checkDecentralandContractsData.js
marcelomorgado: Done.

packages/tasit-sdk/src/scripts/checkDecentralandContractsData.js
marcelomorgado: Done (with a correction too).

packages/tasit-sdk/src/scripts/checkDecentralandContractsData.js
marcelomorgado: Done.

packages/tasit-sdk/src/scripts/checkDecentralandContractsData.js
marcelomorgado: Resolving this.

packages/tasit-contracts/src/scripts/populateDecentralandContracts.js
marcelomorgado: Done.

packages/tasit-contracts/src/scripts/populateDecentralandContracts.js
pcowgill: C) I think part of the reason for fingerprinting estates is because that is possible. But maybe it's disallowed on the front end.

packages/tasit-contracts/src/scripts/populateDecentralandContracts.js
marcelomorgado: Extracted to: https://github.com/tasitlabs/TasitSDK/issues/295

packages/tasit-contracts/src/scripts/populateDecentralandContracts.js
pcowgill: Thanks! Resolving

packages/tasit-sdk/src/scripts/checkDecentralandContractsData.js
pcowgill: Thanks! Resolving

packages/tasit-sdk/src/scripts/checkDecentralandContractsData.js
pcowgill: Thanks! Resolving

packages/tasit-sdk/src/scripts/checkDecentralandContractsData.js
pcowgill: Thanks! Resolving

packages/tasit-contracts/src/scripts/populateDecentralandContracts.js
marcelomorgado: Done.

packages/tasit-sdk/package.json
pcowgill: I don't think all deps are installed on TasitSDK node_modules. Just those with at least 2 packages using them.

https://github.com/lerna/lerna/blob/master/doc/hoist.md

If the Tasit SDK is going to directly use the `node-fetch`, I'd rathre add it back as a dev dependency

packages/tasit-contracts/src/scripts/populateDecentralandContracts.js
marcelomorgado: Oh, I've just copied this function from here https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Math/random#Getting_a_random_integer_between_two_values
Seems that is a way to parse the number to integer.

packages/tasit-contracts/src/scripts/populateDecentralandContracts.js
pcowgill: Thanks. Resolving

packages/tasit-contracts/src/scripts/populateDecentralandContracts.js
pcowgill: Thanks! Resolving

packages/tasit-contracts/src/scripts/populateDecentralandContracts.js
pcowgill: Ah ok - makes sense. Resolving this!

packages/tasit-contracts/src/scripts/populateDecentralandContracts.js
marcelomorgado: Done.

packages/tasit-contracts/src/scripts/populateDecentralandContracts.js
```diff
@@ -78,263 +229,184 @@ const createEstate = async (estateContract, landContract, estate, wallet) => {
...
+  const { parcels: estateParcels } = estate;
+
+  for (let parcel of estateParcels)
+    if (findParcel(parcel, listOfParcels)) return true;
+
+  return false;
+};
+
+const getEstatesFromAPI = async () => {
+  // Note: In the future, we can get that same data from Decentraland contracts to move away from this point of centralization in the future
```
pcowgill:
```suggestion
  // Note: In the future, we can get the same data from Decentraland contracts
  // to move away from this point of centralization
```

packages/tasit-contracts/src/scripts/populateDecentralandContracts.js
pcowgill: Thanks! Resolving

packages/tasit-contracts/src/scripts/populateDecentralandContracts.js
```diff
@@ -78,263 +229,184 @@ const createEstate = async (estateContract, landContract, estate, wallet) => {
...
+    return { metadata, parcels };
+  });
+
+  // Keep estates with small number of parcels to avoid out-of-gas on creation
+  const estates = allEstates.filter(e => e.parcels.length < 10);
+  return estates;
+};
+
+const getParcelsFromAPI = async () => {
+  // Note: In the future, we can get that same data from Decentraland contracts to move away from this point of centralization in the future
```
pcowgill:
```suggestion
  // Note: In the future, we can get the same data from Decentraland contracts
  // to move away from this point of centralization
```

packages/tasit-sdk/src/scripts/checkDecentralandContractsData.js
```diff
@@ -0,0 +1,90 @@
...
+        const image = await fetch(
+          `https://api.decentraland.org/v1/estates/${assetId}/map.png`
+        );
+        const imageData = (await image.buffer()).toString("base64");
+        expect(
+          imageData,
+          `The image of the estate (id: ${assetId}) is blank`
+        ).not.equals(blankImageData);
+      } else if (isParcel) {
+        // Note: Parcels assets always show correct image
```
pcowgill:
```suggestion
        // Note: Parcel assets always show correct image
```

packages/tasit-sdk/src/scripts/checkDecentralandContractsData.js
```diff
@@ -0,0 +1,90 @@
...
+import DecentralandUtils from "../helpers/DecentralandUtils";
+
+const fetch = require("node-fetch");
+
+const { TWO, TEN, BILLION, WEI_PER_ETHER } = constants;
+
+describe("Decentraland App pre-conditions", () => {
+  const mana = new Mana(MANA_ADDRESS);
+  // Note: Since accounts[0] is being used as owner of all deployed contracts
+  // We are using accounts[9] for Gnosis Safe test cases to make sure that it haven't undesired extra privilege access.
```
pcowgill:
```suggestion
  // We are using accounts[9] for Gnosis Safe test cases to make sure that it
  // doesn't have undesired extra privilege access.
```

packages/tasit-sdk/package.json
pcowgill: @marcelomorgado I think this is the only thing gating merging

marcelomorgado: Weird, here even only `tasit-action` having it as a dependency the module is inside of `TasitSDK/node_modules` and the `tasit-sdk` script is working.
Should I back the dependency to `tasit-sdk` anyway?

pcowgill: I wonder if that's due to the last time we ran lerna hoist with a different package.json?

pcowgill: Yeah, I think it's worth adding the dependency back if we'll use it directly for development in that package

marcelomorgado: > I wonder if that's due to the last time we ran lerna hoist with a different package.json?

I didn't get it


> Yeah, I think it's worth adding the dependency back if we'll use it directly for development in that package

Done.

pcowgill: > I didn't get it

If we ran the `lerna --hoist` command most recently before editing the `package.json` file and rerunning the command, then it would make sense that `node-fetch` would still have been hoisted to the root of Tasit SDK.

marcelomorgado: Hmmm, I've run the clean+bootstrap script after removal of `node-fetch` from `tasit-sdk`

pcowgill: Ah ok, then that's probably not the explanation. Does the clean command delete the 1st-party `node_modules` folders from our packages?

marcelomorgado: Yes.

pcowgill: Ah ok, good to know. Feel free to resolve this thread.

marcelomorgado: Resolving this.

PR #345, opened 2019-04-20
Greenkeeper fix - Using greenkeeper-lockfile package

### LINKED ISSUE #244: Fix greenkeeper
Greenkeeper isn't working well. All branches are broken because the tool isn't updating project `package-lock.json`.
I've sent that e-mail:

> Hi,
> We are using lerna to manage our monorepo and using --hoist option to install dependencies of packages. That option installs these on the root level node_modules/ and use the root level package-lock.json for all packages. Seems that setup doesn't work with Greenkeeper since the root level package-lock.json not being updated.
>
> If we use the greenkeeper-lock package that will be solved?
>
> Thanks!
> Marcelo Morgado

And they answered:

> Heya Marcelo,
> 
> We designed the native lockfile support in a way that it can fail silently and only produce a package.json update, which greenkeeper-lockfile then interprets as having to run itself, to update a lock. If greenkeeper-lockfile detects a lockfile update, it will refuse to run itself.
> 
> We’ve made greenkeeper-lockfile accept custom arguments to support various yarn and npm options, but it doesn’t know about lerna specifically.
> 
> If you could share how you need greenkeeper-lockfile to run, we can probably help to get that going.
> 
> Best
> Jan

marcelomorgado:
### Issue link

#244 

### Auto-close the issue?

Closes #244 

### Types of changes

### Notes

Not sure if that will solve our problem but seems a good attempt.
Reference: https://github.com/greenkeeperio/greenkeeper-lockfile#circleci-workflows

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -20,15 +20,15 @@ jobs:
       - checkout
 
       - run:
-          name: install npm
+          name: Installing npm
           command: "sudo npm install -g npm@latest"
 
       - run:
-          name: install greenkeeper-lockfile
+          name: Installing greenkeeper-lockfile
           command: "sudo npm install -g greenkeeper-lockfile@latest"
 
       - run:
-          name: run greenkeeper-lockfile-update
+          name: Running greenkeeper-lockfile-update
           command: "greenkeeper-lockfile-update"
 
       # Download and cache dependencies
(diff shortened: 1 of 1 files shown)


.circleci/config.yml
```diff
@@ -31,15 +39,18 @@ jobs:
             - dependency-cache-
 
       - run:
-          name: install-npm-packages
+          name: Bootstraping packages
```
pcowgill: Bootstrapping

pcowgill:
```suggestion
          name: Bootstrapping packages
```

Issue #253, opened 2019-03-12
TasitAction trigger error without listening to "confirmation" event


marcelomorgado:
Some error (orphan block, failed tx) events are been trigger only from the confirmationListener

marcelomorgado:
It seems an error because we are using `action.waitForNonceToUpdate();` as success path for the action. And an empty `confirmation` listener is needed to get errors.
Without using that, having a `confirmation` event for every action seems natural.
That said, my question is: There is a use case of an action with only an error event?

pcowgill:
> It seems an error because we are using `action.waitForNonceToUpdate();` as success path for the action. And an empty `confirmation` listener is needed to get errors.
> Without using that, having a `confirmation` event for every action seems natural.
> That said, my question is: There is a use case of an action with only an error event?

You mean an action with only an error event listener and no confirmation listener? If that's what you mean, I can't think of a use for that.

Or do you mean an action where there is never a confirmation event but just an error event? If that's what you mean, yeah, I think for instance we want to pass reverts, etc. back through to the error event listener, and that might happen even with 0 confirmations, right?

pcowgill:
> an empty `confirmation` listener is needed to get errors

Why is that? Would it be very complicated to change it so that this isn't true?

marcelomorgado:
1. I've meant an action with only an error event listener and no confirmation listener.
2. To get if the tx was failed (receipt.status) we need to be listening to the transaction to trigger that error.
If we want to support that case, we can auto-subscribe to all actions looking for errors. What do you think?

pcowgill:
1. Okay, gotcha.
2. Makes sense. Yep, I think that's what I was always picturing we would do - I like that idea!

PR #354, opened 2019-04-25
Updating ganache-cli (and catch errors without RPC VMErrors)

### LINKED ISSUE #343: Error listener aren't being trigged without RPC VMError
Since `ganache-cli` 6.4.2 some tests aren't passing because error listeners aren't being trigged with the `--noVMErrorsOnRPCResponse` option (same are passing without that).
Steps to reproduce:
1. Update `ganache-cli` to 6.4.2 or later (tested on 6.4.2 and 6.4.3)
2. Run `npm run boostrap` (TasitSDK repo root)
3. Run `npm test` (`tasit-action` package).

marcelomorgado:
### Issue link

#343 
#253

### Auto-close the issue?

Closes #343

### Types of changes

### DIFF
--- a/packages/tasit-action/src/contract/Contract.test.js
+++ b/packages/tasit-action/src/contract/Contract.test.js
@@ -133,9 +133,16 @@ describe("TasitAction.Contract", () => {
         sampleContract.on("error", errorListener);
 
         action = sampleContract.revertWrite("some string");
+
+        // Some error (orphan block, failed tx) events are being triggered only from the confirmationListener
+        // See more: https://github.com/tasitlabs/TasitSDK/issues/253
+        action.on("confirmation", () => {});
+
         await action.waitForNonceToUpdate();
 
-        expect(errorListener.callCount).to.equal(1);
+        await mineBlocks(provider, 1);
+
+        expect(errorListener.called).to.be.true;
       });
 
       it("and Action error event on action error", async () => {
@@ -146,12 +153,18 @@ describe("TasitAction.Contract", () => {
 
         action = sampleContract.revertWrite("some string");
 
+        // Some error (orphan block, failed tx) events are being triggered only from the confirmationListener
+        // See more: https://github.com/tasitlabs/TasitSDK/issues/253
+        action.on("confirmation", () => {});
+
         action.on("error", actionErrorListener);
 
         await action.waitForNonceToUpdate();
 
-        expect(contractErrorListener.callCount).to.equal(1);
-        expect(actionErrorListener.callCount).to.equal(1);
+        await mineBlocks(provider, 1);
+
+        expect(contractErrorListener.called).to.be.true;
+        expect(actionErrorListener.called).to.be.true;
       });
 
       it("on contract event listener error", async () => {
--- a/packages/tasit-action/src/erc721/ERC721Full.test.js
+++ b/packages/tasit-action/src/erc721/ERC721Full.test.js
@@ -223,8 +223,6 @@ describe("TasitAction.ERC721.ERC721Full", () => {
         [ana.address, SAMPLE_CONTRACT_ADDRESS],
         [1, 0]
       );
-
-      action.off("confirmation");
     });
 
     it("should trigger an error if the user is listening for errors for an action and tries safeTransferFrom to a contract without onERC721Received", async () => {
--- a/packages/tasit-action/src/erc721/ERC721Full.test.js
+++ b/packages/tasit-action/src/erc721/ERC721Full.test.js
@@ -154,7 +154,11 @@ describe("TasitAction.ERC721.ERC721Full", () => {
       expect(event.to).to.equal(bob.address);
       expect(event.tokenId.toNumber()).to.equal(tokenId);
 
-      await expectExactTokenBalances(erc721, [ana.address, bob.address], [0, 1]);
+      await expectExactTokenBalances(
+        erc721,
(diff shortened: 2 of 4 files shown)


packages/tasit-action/src/contract/Contract.test.js
```diff
@@ -146,12 +153,18 @@ describe("TasitAction.Contract", () => {
...
         action.on("error", actionErrorListener);
 
         await action.waitForNonceToUpdate();
 
-        expect(contractErrorListener.callCount).to.equal(1);
-        expect(actionErrorListener.callCount).to.equal(1);
+        await mineBlocks(provider, 1);
+
+        expect(contractErrorListener.called).to.be.true;
+        expect(actionErrorListener.called).to.be.true;
```
pcowgill: Any reason you're moving away from `callCount`?

packages/tasit-action/src/erc721/ERC721Full.test.js
```diff
@@ -188,9 +196,7 @@ describe("TasitAction.ERC721.ERC721Full", () => {
 
       const contractErrorListener = message => {
```
pcowgill: Maybe we should call this param something other than `message` because the `error` within `message` contains `message`, which is confusing

packages/tasit-action/src/erc721/ERC721Full.test.js
```diff
@@ -219,9 +232,7 @@ describe("TasitAction.ERC721.ERC721Full", () => {
 
       const actionErrorListener = message => {
         const { error } = message;
-        expect(error.message).to.equal(
-          "Action with error: VM Exception while processing transaction: revert"
-        );
+        expect(error.message).to.equal("Action failed.");
```
pcowgill: Should we extract message from error here using destructuring assignment to be consistent?

packages/tasit-action/src/erc721/ERC721Full.test.js
marcelomorgado: Should we change the API from error listeners?


```javascript
// Subscription._emitErrorEvent(error):
// from
    const message = { error };
    errorEventListener.listener(message);
// to
    errorEventListener.listener(error);
```

packages/tasit-action/src/contract/Contract.test.js
marcelomorgado: Because the number of calls is being non-deterministic, I can move to `.to.be.at.least(1)` and make a comment about that, should I?

pcowgill: Yeah, I think I prefer that if that works for you

pcowgill: And the comment would definitely be helpful

packages/tasit-action/src/erc721/ERC721Full.test.js
pcowgill: Hmm, interesting. Yeah, I'd be open to that if you think that's better. Can you think of any other information we might want to decorate the object with that wouldn't fit in the error object?

What if we wanted to give our own name for the error but still pass through the original error name? Is our current approach better for that?

packages/tasit-action/src/contract/Contract.test.js
marcelomorgado: I think that it was addresses by the https://github.com/tasitlabs/TasitSDK/pull/358

packages/tasit-action/src/erc721/ERC721Full.test.js
marcelomorgado: I think that we can extend the original error if we want to add more info to that sometime.
I've written a sample: https://es6console.com/jv3m7125/

pcowgill: Yeah, that's perfect

pcowgill: Feel free to resolve this and create an issue for that?

The idea of renaming the param, which is the original subject of this thread, can be resolved

packages/tasit-action/src/contract/Contract.test.js
pcowgill: Yep. Thanks. Resolving this.

packages/tasit-action/src/erc721/ERC721Full.test.js
marcelomorgado: > Feel free to resolve this and create an issue for that?
It was partially solved w/ the PR https://github.com/tasitlabs/TasitSDK/pull/369.
Missing using inherit to decorate the error object. Extracted to: https://github.com/tasitlabs/TasitSDK/issues/370
Resolving this.

packages/tasit-action/src/erc721/ERC721Full.test.js
marcelomorgado: Done for all cases on: https://github.com/tasitlabs/TasitSDK/issues/370

packages/tasit-action/src/erc721/ERC721Full.test.js
pcowgill: Perfect, thanks!

packages/tasit-action/src/erc721/ERC721Full.test.js
pcowgill: Thanks! Resolving this

Issue #219, opened 2019-02-25
Test suites should captures how we expect the library to be used


marcelomorgado:
On UX context, our users will dispatch actions that will trigger some UI response for success or failure of the processes.
For now, all of our test suites wasn't written closer that way, we are using in almost all of test cases the `waitForNonceToUpdate` function and we expect that will not be used for our users.
The way to write test cases to get closer as we expected that the code will be it to using `mocha` `done()` function.
Refs: 
https://github.com/tasitlabs/TasitSDK/pull/207#discussion_r258306552
https://github.com/mochajs/mocha/issues/2988#issuecomment-327200194
https://mochajs.org/#asynchronous-code

marcelomorgado:
I'll start to use `done()` from now for new test cases that I'll wrote.

marcelomorgado:
Hey @pcowgill, I'm working on that on gnosis safe test suites, what do you think?

**Before** (with waiting)
```javascript
    it("wallet owner should withdraw some ethers", async () => {
      const signers = [anaWallet];
      const { address: toAddress } = anaWallet;
      const value = ONE;

      gnosisSafe.setWallet(anaWallet);
      const execTxAction = await gnosisSafe.transferEther(
        signers,
        toAddress,
        value
      );
      await execTxAction.waitForNonceToUpdate();

      const balanceAfterWithdraw = await provider.getBalance(
        GNOSIS_SAFE_ADDRESS
      );
      expect(`${balanceAfterWithdraw}`).to.equal(`${ZERO}`);
    });
```

**After** (without waiting)
```javascript
    it("wallet owner should withdraw some ethers", async () => {
      const signers = [anaWallet];
      const { address: toAddress } = anaWallet;
      const value = ONE;

      gnosisSafe.setWallet(anaWallet);
      const action = await gnosisSafe.transferEther(signers, toAddress, value);

      const onConfirmation = async message => {
        const { data } = message;
        const { args } = data;

        action.unsubscribe();

        const balance = await provider.getBalance(GNOSIS_SAFE_ADDRESS);
        expect(`${balance}`).to.equal(`${ZERO}`);
      };

      const onError = message => {
        const { error } = message;
        console.log(error);
      };

      await dispatch(action, onConfirmation, onError);
    });

    const dispatch = async (action, onConfirmation, onError) => {
      return new Promise((resolve, reject) => {
        action.on("confirmation", async message => {
          try {
            await onConfirmation(message);
            resolve();
[...]

Issue #187, opened 2019-02-02
Add support for gas param creation in the SDK so that it is abstracted away when calling contract functions


pcowgill:

pcowgill:
The the Decentraland integration test would change so that `executeOrder` isn't called with `gasParams`

Issue #118, opened 2019-01-16
TasitAccount main object


marcelomorgado:
For now, `tasit-account` creates a ethers.js wallet object. Should we change that? 
Could meta-tx support demands more fields to be used here?
E.g.:
```javascript
// TasitAcount.create()
// > Acount { wallet: ..., metaTxInfos..., etc }
```

Refs: https://github.com/tasitlabs/TasitSDK/pull/71#discussion_r243742850

PR #357, opened 2019-04-26
Action.send();

### LINKED ISSUE #162: Allows Action to send tx after object creation
Today it works like this:
```javascript
action = contract.writeFunction(); // tx broadcast
action.on("error", errorListener);
```

Should be
```javascript
action = contract.writeFunction(); 
action.on("error", errorListener);
action.send(); // tx broadcast
```

Refs: https://github.com/tasitlabs/TasitSDK/pull/159#discussion_r250767004

marcelomorgado:
### Issue link

#162 

### Auto-close the issue?

Closes #162 

### Types of changes

Breaking change (fix or feature that would cause existing functionality not to work as expected)

### Notes

- The `send()` function is async and the use of `await` on the call is necessary for now (Since we are still using `waitForOneConfirmation()`), it should be changed on the https://github.com/tasitlabs/TasitSDK/issues/219

- The way that `GnosisSafe` was coded introduced an extra complexity to the `Action` object: A function that should be sync but creates an action from a sequence of async calls (See: `GnosisSafe.executeTransaction()`). The solution that I've found is assuming that the `rawTx` param (from Action constructor) could be a promise or not.
Maybe a better name could be used instead of `rawTx` (maybe `rawAction`?).

### DIFF
--- a/packages/tasit-action/src/contract/Action.js
+++ b/packages/tasit-action/src/contract/Action.js
@@ -6,38 +6,50 @@ import ConfigLoader from "../ConfigLoader";
 //  and/or MetaTxAction subclasses
 export class Action extends Subscription {
   #provider;
-  #tx; // deprecated
-  #signedTx;
+  #signer;
+  #rawTx;
+  #tx;
   #txConfirmations;
   #timeout;
   #lastConfirmationTime;
 
-  constructor(signedTxPromise, provider) {
+  constructor(rawTx, signer) {
     // Provider implements EventEmitter API and it's enough
     //  to handle with transactions events
-    super(provider);
+    super(signer.provider);
 
     const { events } = ConfigLoader.getConfig();
     const { timeout } = events;
 
-    // this.#tx = txPromise.then(
-    //   tx => {
-    //     return tx;
-    //   },
-    //   error => {
-    //     this._emitErrorEvent(new Error(`Action with error: ${error.message}`));
-    //   }
-    // );
-    this.#signedTx = signedTxPromise;
-
+    this.#rawTx = rawTx;
+    this.#signer = signer;
     this.#timeout = timeout;
-    this.#provider = provider;
+    this.#provider = signer.provider;
     this.#txConfirmations = 0;
   }
 
+  #signAndSend = async () => {
+    const gasLimit = await this.#provider.estimateGas(this.#rawTx);
+    const nonce = await this.#provider.getTransactionCount(
+      this.#signer.address
+    );
+
+    this.#rawTx = { ...this.#rawTx, nonce, gasLimit };
+
+    const signedTx = await this.#signer.sign(this.#rawTx);
+
+    this.#tx = await this.#provider.sendTransaction(signedTx).then(
+      tx => {
+        return tx;
+      },
+      error => {
+        this._emitErrorEvent(new Error(`Action with error: ${error.message}`));
+      }
+    );
+  };
+
   send = async () => {
-    const signedTx = await this.#signedTx;
-    this.#tx = await this.#provider.sendTransaction(signedTx);
+    await this.#signAndSend();
   };
 
   on = (eventName, listener) => {
--- a/packages/tasit-action/src/contract/Action.js
+++ b/packages/tasit-action/src/contract/Action.js
@@ -28,8 +28,8 @@ export class Action extends Subscription {
     this.#txConfirmations = 0;
   }
 
-  getRawTx = () => {
-    return this.#rawTx;
+  _toRaw = async () => {
+    return await this.#rawTx;
   };
 
   #signAndSend = async () => {
--- a/packages/tasit-action/src/contract/Action.js
+++ b/packages/tasit-action/src/contract/Action.js
@@ -29,7 +29,11 @@ export class Action extends Subscription {
   }
 
   #signAndSend = async () => {
-    const gasLimit = await this.#provider.estimateGas(this.#rawTx);
+    // Note: Using hardcoded gas limit because using estimations isn't working right now
+    // See more: https://github.com/tasitlabs/TasitSDK/issues/173
+    //const gasLimit = await this.#provider.estimateGas(this.#rawTx);
+    const gasLimit = 7e6;
+
     const nonce = await this.#provider.getTransactionCount(
       this.#signer.address
     );
--- a/packages/tasit-action/src/contract/Action.js
+++ b/packages/tasit-action/src/contract/Action.js
@@ -54,14 +54,11 @@ export class Action extends Subscription {
 
     const signedTx = await this.#signer.sign(rawTx);
 
-    this.#tx = await this.#provider.sendTransaction(signedTx).then(
-      tx => {
-        return tx;
-      },
-      error => {
-        this._emitErrorEvent(new Error(`Action with error: ${error.message}`));
-      }
-    );
+    try {
+      this.#tx = await this.#provider.sendTransaction(signedTx);
+    } catch (error) {
+      this._emitErrorEvent(new Error(`Action with error: ${error.message}`));
+    }
   };
 
   send = async () => {
--- a/packages/tasit-action/src/contract/Action.js
+++ b/packages/tasit-action/src/contract/Action.js
@@ -28,6 +28,10 @@ export class Action extends Subscription {
     this.#txConfirmations = 0;
   }
 
+  getRawTx = () => {
+    return this.#rawTx;
+  };
+
   #signAndSend = async () => {
     // TODO: Go deep on gas handling.
     // Without that, VM returns a revert error instead of out of gas error.
@@ -44,9 +48,11 @@ export class Action extends Subscription {
       this.#signer.address
     );
 
-    this.#rawTx = { ...this.#rawTx, nonce, ...gasParams };
+    let rawTx = await this.#rawTx;
 
-    const signedTx = await this.#signer.sign(this.#rawTx);
+    rawTx = { ...rawTx, nonce, ...gasParams };
+
+    const signedTx = await this.#signer.sign(rawTx);
 
     this.#tx = await this.#provider.sendTransaction(signedTx).then(
       tx => {
--- a/packages/tasit-action/src/contract/Action.js
+++ b/packages/tasit-action/src/contract/Action.js
@@ -6,12 +6,13 @@ import ConfigLoader from "../ConfigLoader";
 //  and/or MetaTxAction subclasses
 export class Action extends Subscription {
   #provider;
-  #tx;
+  #tx; // deprecated
+  #signedTx;
   #txConfirmations;
   #timeout;
   #lastConfirmationTime;
 
-  constructor(txPromise, provider) {
+  constructor(signedTxPromise, provider) {
     // Provider implements EventEmitter API and it's enough
     //  to handle with transactions events
     super(provider);
@@ -19,19 +20,25 @@ export class Action extends Subscription {
     const { events } = ConfigLoader.getConfig();
     const { timeout } = events;
 
-    this.#tx = txPromise.then(
-      tx => {
-        return tx;
-      },
-      error => {
-        this._emitErrorEvent(new Error(`Action with error: ${error.message}`));
-      }
-    );
+    // this.#tx = txPromise.then(
+    //   tx => {
+    //     return tx;
+    //   },
+    //   error => {
+    //     this._emitErrorEvent(new Error(`Action with error: ${error.message}`));
+    //   }
+    // );
+    this.#signedTx = signedTxPromise;
 
     this.#timeout = timeout;
     this.#provider = provider;
     this.#txConfirmations = 0;
   }
+
+  send = async () => {
+    const signedTx = await this.#signedTx;
+    this.#tx = await this.#provider.sendTransaction(signedTx);
+  };
 
   on = (eventName, listener) => {
     this.#addListener(eventName, listener, false);
--- a/packages/tasit-action/src/contract/Action.js
+++ b/packages/tasit-action/src/contract/Action.js
@@ -13,10 +13,10 @@ export class Action extends Subscription {
   #timeout;
   #lastConfirmationTime;
 
-  constructor(rawTx, signer) {
+  constructor(rawTx, provider, signer) {
     // Provider implements EventEmitter API and it's enough
     //  to handle with transactions events
-    super(signer.provider);
+    super(provider);
 
     const { events } = ConfigLoader.getConfig();
     const { timeout } = events;
@@ -24,21 +24,27 @@ export class Action extends Subscription {
     this.#rawTx = rawTx;
     this.#signer = signer;
     this.#timeout = timeout;
-    this.#provider = signer.provider;
+    this.#provider = provider;
     this.#txConfirmations = 0;
   }
 
   #signAndSend = async () => {
-    // Note: Using hardcoded gas limit because using estimations isn't working right now
-    // See more: https://github.com/tasitlabs/TasitSDK/issues/173
-    //const gasLimit = await this.#provider.estimateGas(this.#rawTx);
-    const gasLimit = 7e6;
+    // TODO: Go deep on gas handling.
+    // Without that, VM returns a revert error instead of out of gas error.
+    // See: https://github.com/tasitlabs/TasitSDK/issues/173
+    //
+    // This command isn't enough
+    // const gasLimit = await this.#provider.estimateGas(this.#rawTx);
+    const gasParams = {
+      gasLimit: 7e6,
+      gasPrice: 1e9,
+    };
 
     const nonce = await this.#provider.getTransactionCount(
       this.#signer.address
     );
 
-    this.#rawTx = { ...this.#rawTx, nonce, gasLimit };
+    this.#rawTx = { ...this.#rawTx, nonce, ...gasParams };
 
     const signedTx = await this.#signer.sign(this.#rawTx);
 
--- a/packages/tasit-action/src/contract/Contract.js
+++ b/packages/tasit-action/src/contract/Contract.js
@@ -163,9 +163,24 @@ export class Contract extends Subscription {
       if (!Utils.isEthersJsSigner(this.#ethersContract.signer))
         throw new Error(`Cannot write data to a Contract without a wallet`);
 
-      const tx = this.#ethersContract[f.name](...args);
+      const data = this.#ethersContract.interface.functions[f.name].encode(
+        args
+      );
 
-      const action = new Action(tx, this.#provider);
+      const signedTx = async () => {
+        const { signer, address } = this.#ethersContract;
+        const rawTx = {
+          gasLimit: 64000,
+          to: address,
+          data: data,
+          nonce: await provider.getTransactionCount(signer.address),
+        };
+        return await signer.sign(rawTx);
+      };
+
+      //const tx = this.#ethersContract[f.name](...args);
+
+      const action = new Action(signedTx(), this.#provider);
 
       const errorListener = message => {
         const { error } = message;
--- a/packages/tasit-action/src/contract/Contract.js
+++ b/packages/tasit-action/src/contract/Contract.js
@@ -170,7 +170,7 @@ export class Contract extends Subscription {
 
       const rawTx = { to, data };
 
-      const action = new Action(rawTx, signer);
+      const action = new Action(rawTx, this.#provider, signer);
 
       const errorListener = message => {
         const { error } = message;
--- a/packages/tasit-action/src/contract/Contract.js
+++ b/packages/tasit-action/src/contract/Contract.js
@@ -163,24 +163,14 @@ export class Contract extends Subscription {
       if (!Utils.isEthersJsSigner(this.#ethersContract.signer))
         throw new Error(`Cannot write data to a Contract without a wallet`);
 
+      const { signer, address: to } = this.#ethersContract;
       const data = this.#ethersContract.interface.functions[f.name].encode(
         args
       );
 
-      const signedTx = async () => {
-        const { signer, address } = this.#ethersContract;
-        const rawTx = {
-          gasLimit: 64000,
-          to: address,
-          data: data,
-          nonce: await provider.getTransactionCount(signer.address),
-        };
-        return await signer.sign(rawTx);
-      };
+      const rawTx = { to, data };
 
-      //const tx = this.#ethersContract[f.name](...args);
-
-      const action = new Action(signedTx(), this.#provider);
+      const action = new Action(rawTx, signer);
 
       const errorListener = message => {
         const { error } = message;
--- a/packages/tasit-action/src/contract/Contract.js
+++ b/packages/tasit-action/src/contract/Contract.js
@@ -168,12 +168,12 @@ export class Contract extends Subscription {
       if (!Utils.isEthersJsSigner(this.#ethersContract.signer))
         throw new Error(`Cannot write data to a Contract without a wallet`);
 
-      const { signer, address: to } = this.#ethersContract;
+      const { signer, address } = this.#ethersContract;
       const data = this.#ethersContract.interface.functions[f.name].encode(
         args
       );
 
-      const rawTx = { to, data };
+      const rawTx = { to: address, data };
 
       const action = new Action(rawTx, this.#provider, signer);
 
--- a/packages/tasit-action/src/contract/Contract.js
+++ b/packages/tasit-action/src/contract/Contract.js
@@ -153,7 +153,7 @@ export class Contract extends Subscription {
 
   #attachReadFunction = f => {
     this[f.name] = async (...args) => {
-      const value = await this.#ethersContract[f.name].apply(null, args);
+      const value = await this.#ethersContract[f.name](...args);
       return value;
     };
   };
@@ -163,7 +163,7 @@ export class Contract extends Subscription {
       if (!Utils.isEthersJsSigner(this.#ethersContract.signer))
         throw new Error(`Cannot write data to a Contract without a wallet`);
 
-      const tx = this.#ethersContract[f.name].apply(null, args);
+      const tx = this.#ethersContract[f.name](...args);
 
       const action = new Action(tx, this.#provider);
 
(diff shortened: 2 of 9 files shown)


pcowgill:
> The send() function is async and the use of await on the call is necessary for now (Since we are still using waitForOneConfirmation()), it should be changed on the #219

Makes sense

> The way that GnosisSafe was coded introduced an extra complexity to the Action object: A function that should be sync but creates an action from a sequence of async calls (See: GnosisSafe.executeTransaction()). The solution that I've found is assuming that the rawTx param (from Action constructor) could be a promise or not.

Makes sense! Thanks for this info, that's helpful.

> Maybe a better name could be used instead of rawTx (maybe rawAction?).

Sounds good

packages/tasit-action/src/contract/Action.js
```diff
@@ -6,32 +6,67 @@ import ConfigLoader from "../ConfigLoader";
 //  and/or MetaTxAction subclasses
 export class Action extends Subscription {
   #provider;
+  #signer;
```
pcowgill: How does this end up being used?

packages/tasit-action/src/contract/Action.js
```diff
@@ -6,32 +6,67 @@ import ConfigLoader from "../ConfigLoader";
...
-    this.#tx = txPromise.then(
+    this.#rawTx = rawTx;
+    this.#signer = signer;
+    this.#timeout = timeout;
+    this.#provider = provider;
+    this.#txConfirmations = 0;
+  }
+
+  _toRaw = async () => {
+    return await this.#rawTx;
```
pcowgill: Can we handle whether to await it or not upstream of this function somehow? Like only call _toRaw once we know the promise has resolved? Or only use this in `tasit-contract-based-account`?

packages/tasit-action/src/contract/Action.js
```diff
@@ -6,32 +6,67 @@ import ConfigLoader from "../ConfigLoader";
...
+      this.#signer.address
+    );
+
+    let rawTx = await this.#rawTx;
+
+    rawTx = { ...rawTx, nonce, ...gasParams };
+
+    const signedTx = await this.#signer.sign(rawTx);
+
+    this.#tx = await this.#provider.sendTransaction(signedTx).then(
```
pcowgill: Any way we can avoid using `.then` here and use another `await` instead?

packages/tasit-action/src/contract/Contract.js
```diff
@@ -51,6 +51,11 @@ export class Contract extends Subscription {
     this.#addFunctionsToContract();
   };
 
+  getWallet = () => {
+    const { signer: wallet } = this.#ethersContract;
+    return wallet;
```
pcowgill: Is it a "wallet" or an "account"? I think it's only a "wallet" when an "account" or "contract" is wrapped in an app.

packages/tasit-action/src/contract/Contract.js
```diff
@@ -163,9 +168,14 @@ export class Contract extends Subscription {
       if (!Utils.isEthersJsSigner(this.#ethersContract.signer))
         throw new Error(`Cannot write data to a Contract without a wallet`);
 
-      const tx = this.#ethersContract[f.name].apply(null, args);
+      const { signer, address: to } = this.#ethersContract;
+      const data = this.#ethersContract.interface.functions[f.name].encode(
+        args
+      );
+
+      const rawTx = { to, data };
```
pcowgill: I think it might be easier to read if it said "to: address" here instead of renaming address to to a few lines up. Because I wondered a few lines up why it was being called to, but once I got here it made sense.

packages/tasit-action/src/erc721/ERC721Full.test.js
```diff
@@ -163,12 +165,15 @@ describe("TasitAction.ERC721.ERC721Full", () => {
 
     it("should transfer an approved token", async () => {
       erc721 = new ERC721Full(ERC721_FULL_ADDRESS, ana);
+      erc721.setWallet(ana);
```
pcowgill: Why wasn't this needed before?

packages/tasit-contract-based-account/src/GnosisSafe.js
```diff
@@ -11,14 +11,6 @@ const { abi: gnosisSafeABI } = GnosisSafeInfo;
 const { abi: erc20ABI } = MyERC20Full;
 const { abi: erc721ABI } = MyERC721Full;
 
-// TODO: Go deep on gas handling.
-// Without that, VM returns a revert error instead of out of gas error.
-// See: https://github.com/tasitlabs/TasitSDK/issues/173
-const gasParams = {
```
pcowgill: We might want to use higher gas params for the Gnosis Safe tests than we need to for the regular `tasit-action` tests

packages/tasit-action/src/contract/Action.js
pcowgill: I guess it's often enough that the signer isn't the sender in our lib with contract-based accounts and meta-tx that we shouldn't default this like we used to.

packages/tasit-contract-based-account/src/GnosisSafe.js
```diff
@@ -229,10 +223,11 @@ export default class GnosisSafe extends Contract {
       gasPrice,
       gasToken,
       refundReceiver,
-      signatures,
-      gasParams
+      signatures
     );
 
-    return execTxAction.getTransaction();
+    const rawTx = await execTxAction._toRaw();
```
pcowgill: Yeah, I think maybe _toRaw could just be defined for `tasit-contract-based-account`

packages/tasit-contracts/src/scripts/populateDecentralandContracts.js
```diff
@@ -91,11 +91,8 @@ const gnosisSafeContract = new GnosisSafe(GNOSIS_SAFE_ADDRESS);
 const provider = ProviderFactory.getProvider();
 
 const cancelSellOrder = async (nftAddress, id) => {
-  const action = marketplaceContract.cancelOrder(
-    nftAddress,
-    `${id}`,
-    gasParams
```
pcowgill: I really like that this (`gasParams`) is gone from this script now.

pcowgill:
@marcelomorgado What does `TBD:` mean in the name of this PR?

marcelomorgado:
> @marcelomorgado What does `TBD:` mean in the name of this PR?

To Be Discussed

packages/tasit-action/src/contract/Action.js
marcelomorgado: Hmm, the `Action` is receiving the `rawTx` and the `signer`, on the `send()` function (it's an alias for `signAndSend()` the `rawTx` is being decorated (gasParam and nonce), sign and sent.
We can have the action receiving the `signedTx` instead but will need to handle with gas and nonce outside?
A question: Shouldn't all transactions (even the meta-tx) be signed before broadcasted?

packages/tasit-action/src/contract/Action.js
marcelomorgado: 1. I need to think about it, not sure.
2. For now, we are using that function only on `tasit-contract-based-account` GnosisSafe class.

packages/tasit-action/src/contract/Action.js
marcelomorgado: Sure, I've just realized that `Promise(resolve,request)` is the same as `try { await } catch { }`.
Done.

packages/tasit-action/src/contract/Contract.js
marcelomorgado: I thought that since the `tasit-account` is creating an `ethers.js Wallet` we should use this name for now, and when we wrap that with an `Account` object we'll name it as `account`.

Refs:
```javascript
  // Note: For now, `tasit-account` creates a ethers.js wallet object
  // If that changes, maybe this method could be renamed to setAccount()
  setWallet = wallet => {
```

packages/tasit-action/src/contract/Contract.js
marcelomorgado: Done.

packages/tasit-action/src/erc721/ERC721Full.test.js
marcelomorgado: Oh, good catch.
I've added this line for debugging and forgot to remove.
Done.

packages/tasit-contract-based-account/src/GnosisSafe.js
marcelomorgado: Yes, I'm using these (higher) params as default now but I think that we need to work on estimation. Maybe as the next issue?

pcowgill:
> > @marcelomorgado What does `TBD:` mean in the name of this PR?
> 
> To Be Discussed

Ah ok cool. It's often used to mean `To Be Determined`, so I was confused

packages/tasit-contract-based-account/src/GnosisSafe.js
marcelomorgado: Any idea how I can do that?
`GnosisSafe` extends `Contract`, `Action` is a separated object.

packages/tasit-contracts/src/scripts/populateDecentralandContracts.js
marcelomorgado: Yes, me too. Now we can handle gas just inside of the `Action`.

packages/tasit-contract-based-account/src/GnosisSafe.js
pcowgill: Yep, I think working on estimation can make sense for an issue pretty soon.

What I'm advocating for here is that we use these higher params as a default everywhere except for the tests in `tasit-action` itself, where we know `gasParams` can be lower. So `tasit-action` could override the default in its own tests?

packages/tasit-action/src/erc721/ERC721Full.test.js
pcowgill: Ah ok, glad we caught this. Whenever I add debugging lines I like to add `// TODO: Remove me before merging`

pcowgill: Resolving this

packages/tasit-action/src/contract/Contract.js
pcowgill: Thanks! Resolving

packages/tasit-action/src/contract/Contract.js
pcowgill: That's a fair point.

I might dispute ethers.js calling it a "wallet", but I think when an account is wrapped in even a js object it's already something disctinct, so maybe it is a wallet at that point by my own definition.

Okay, let's keep this as-is

pcowgill: Hmmm, although Etherum account wrapped in ethers.js wallet wrapped in Tasit "account" is probably too confusing.

packages/tasit-contract-based-account/src/GnosisSafe.js
marcelomorgado: For sure, we can do that.
I've thought that will be desirable not exposing gas setup thru our API and having an excellent gas estimation solution.

packages/tasit-action/src/contract/Action.js
pcowgill: Oh, yep!

Great, thanks. Resolving.

packages/tasit-action/src/contract/Action.js
pcowgill: 1. Okay. Keep me posted. My intuition tells me that move would make sense, but I haven't thought it through fully.
2. Yep, I thought so. That's part of my rough reasoning for this.

packages/tasit-action/src/contract/Action.js
```diff
@@ -6,32 +6,67 @@ import ConfigLoader from "../ConfigLoader";
...
   #provider;
+  #signer;
+  #rawTx;
   #tx;
   #txConfirmations;
   #timeout;
   #lastConfirmationTime;
 
-  constructor(txPromise, provider) {
+  constructor(rawTx, provider, signer) {
```
pcowgill: This is the line I should have commented on. Moving this https://github.com/tasitlabs/TasitSDK/pull/357/files/ea53439d35f7f5e022c6d3e1e4f09d6b4b02b617#r279085297 thread here.

packages/tasit-action/src/contract/Action.js
pcowgill: Moving thread here https://github.com/tasitlabs/TasitSDK/pull/357/files/ea53439d35f7f5e022c6d3e1e4f09d6b4b02b617#r279128314

packages/tasit-action/src/contract/Action.js
pcowgill: > Hmm, the `Action` is receiving the `rawTx` and the `signer`, on the `send()` function (it's an alias for `signAndSend()` the `rawTx` is being decorated (gasParam and nonce), sign and sent.

That makes sense that it's signed after gasParam and nonce, but I'm glad you reminded me of that in this context.

> We can have the action receiving the `signedTx` instead but will need to handle with gas and nonce outside?

Hmm, I wasn't thinking of moving gas and nonce outside. I was thinking of assuming who the signer is whenever possible outside the scope of action creation. So maybe already setting it for the `tasit-account` and then the action uses that. Half-baked thought, though.

> A question: Shouldn't all transactions (even the meta-tx) be signed before broadcasted?

Yes, they definitely need to be signed by an account before being broadcast to a relayer or the Ethereum network. I guess I could imagine an app powered the Tasit SDK building an action that is then serialized and deep-linked over to a mobile wallet app that then broadcasts it to a relayer or the Ethereum network, though. But we can build around such a feature when the time comes, not preemptively

packages/tasit-contract-based-account/src/GnosisSafe.js
pcowgill: I agree that requiring gas setup thru our API is not desirable. Exposing it as optional with a default might be okay though - as long as this isn't in a very prominent, distracting place in the docs

packages/tasit-contracts/src/scripts/populateDecentralandContracts.js
pcowgill: Yep, cool! Resolving this

packages/tasit-contract-based-account/src/GnosisSafe.js
pcowgill: Good point. I guess the underscore does a good enough job of showing that it only will be used internally

pcowgill: Oh, maybe there could be something separate in `tasit-contract-based-account` that extends `Action` and adds this functionality? Requires more thought on my end, though.

packages/tasit-action/src/contract/Action.js
pcowgill: Resolving this since the thread was moved

pcowgill:
@marcelomorgado I don't think any of the open comment threads gate merging this. Nicely done with this PR, by the way! Any objection to merging it now?

packages/tasit-contract-based-account/src/GnosisSafe.js
marcelomorgado: Since es6 hasn't properly protected methods, I think that will be almost the same, we'll need to expose the `rawTx` field (maybe as _rawTx) instead of having that thru the function.

marcelomorgado:
> @marcelomorgado I don't think any of the open comment threads gate merging this. Nicely done with this PR, by the way! Any objection to merging it now?

Let's merge!

packages/tasit-contract-based-account/src/GnosisSafe.js
pcowgill: Ah yeah - never mind! Getter function is better

packages/tasit-action/src/contract/Action.js
marcelomorgado: I think we can move the signer from the constructor to a setter function and check it that's assigned on the `send` function.

pcowgill: Would that mean that we'd have to set the signer for every action as a separate step after building it? I think we'd only want to go that route if we could use that setter on `tasit-account` and then actions would use that signer.

packages/tasit-action/src/contract/Action.js
marcelomorgado: I've made a change on this PR https://github.com/tasitlabs/TasitSDK/pull/360
On these changes, we are assuming that the `Action` will expect a sync `rawTx` on the constructor but the `signAndSend` function stills using `await` to resolve that. It isn't 100% solution, though.
Another path is to force `rawTx` to be sync (removing the `await` inside of the `signAndSend` but that'll make `GnosisSafe.executeTransaction` to be async (on the way that is written for now).

packages/tasit-action/src/contract/Contract.js
marcelomorgado: Should we keep calling (the future implementation) as tasit account?

packages/tasit-contract-based-account/src/GnosisSafe.js
marcelomorgado: Ok, how about the API?
For now we're almost always using the `Action` thru the `Contract` object.
Should we expose the gas customization for each contract write contract (as ethers does):
```javascript
contract.callWrite(arg1, arg2, gasParams);
```

Or having a setter for the contract?

```javascript
contract.gasParams(gasParams);
contract.callWrite(arg1, arg2);
```

packages/tasit-contract-based-account/src/GnosisSafe.js
marcelomorgado: Okay, could I resolve this?

packages/tasit-action/src/contract/Action.js
marcelomorgado: Hmm, this change (signer from constructor to setter) will be a tiny change for now, since we are using `Contract` to create actions, this change will be almost transparent. But could reflect better how the `Action` behavior should be (Allows creation of an Action to be signed later).

Note: The current `Contract` code will throw an error if a write function be called without a signer/wallet set. Maybe we should change that check to the `Action.send` function.

Could you clarify that sentence? 
>  if we could use that setter on tasit-account and then actions would use that signer.

packages/tasit-action/src/contract/Action.js
```diff
@@ -6,33 +6,65 @@ import ConfigLoader from "../ConfigLoader";
...
```
marcelomorgado: Creating a thread for that:

> > Maybe a better name could be used instead of rawTx (maybe rawAction?).

> Sounds good

To track easily.

marcelomorgado: https://github.com/tasitlabs/TasitSDK/pull/360

packages/tasit-action/src/contract/Contract.js
pcowgill: I think so. We don't want people to think of it as a "wallet" per se from a product perspective

packages/tasit-contract-based-account/src/GnosisSafe.js
pcowgill: What do you mean by `contract write contract`?

pcowgill: I could imagine a contract having two write functions with very different gas needs, so allowing custom ones for each function seems to make sense (although spending time on this right now is a low priority - we can track it in an issue)

packages/tasit-contract-based-account/src/GnosisSafe.js
pcowgill: Yep. I'm resolving it now.

packages/tasit-action/src/contract/Action.js
pcowgill: That all sounds good to me.

>> if we could use that setter on tasit-account and then actions would use that signer.

>Could you clarify that sentence?

I guess I actually meant using the `tasit-account` account as the signer in `Contract` within `tasit-action`

packages/tasit-action/src/contract/Action.js
pcowgill: Sounds good, thanks

packages/tasit-action/src/contract/Action.js
pcowgill: I don't have time to think this through fully right now - feel free to go with whatever you thinks makes the most sense for now, and I'll circle back to this soon

packages/tasit-contract-based-account/src/GnosisSafe.js
marcelomorgado: > What do you mean by `contract write contract`?

Sorry: contract write function

marcelomorgado: We already have an issue about that: https://github.com/tasitlabs/TasitSDK/issues/187

pcowgill: > Sorry: contract write function

Ah ok, makes sense That's what I assumed in my comment, so that comment sitll stands.

pcowgill: > We already have an issue about that: #187

But that issue is mostly about allowing for not specifying gas params outside the SDK

Whereas what we're discussing here is letting users optionally specify gas params outside the SDK if they want. Right? Should we mention that in that issue and broaden its scope, or create a new issue?

packages/tasit-action/src/contract/Action.js
pcowgill: The approach you're using in the latest version of the code works for me.

Just to make sure I understand, why would removing the `await` inside `signAndSend` mean that `GnosisSafe.executeTransaction` would be async?

marcelomorgado: If we remove `await` from the `signAndSend` the `Action` class won't accept `promise-of-rawTx` as a constructor argument.
That way is how I've found to keep `GnosisSafe` action-create-functions as sync (as `Contract` does). ([ref](https://github.com/tasitlabs/TasitSDK/blob/4111170669afb4447cb95315f45fd4ecd1fa02ec/packages/tasit-contract-based-account/src/GnosisSafe.js#L148))
If `Action` stop to accept `txRaw` as a promise, I'm not seeing a way to keep `GnosisSafe.executeTransaction` as sync right now.

packages/tasit-action/src/contract/Contract.js
marcelomorgado: Related to https://github.com/tasitlabs/TasitSDK/issues/118
Could we close this thread?

packages/tasit-contract-based-account/src/GnosisSafe.js
marcelomorgado: I think a specific issue will be better. Extracted to: https://github.com/tasitlabs/TasitSDK/issues/371

packages/tasit-action/src/contract/Action.js
marcelomorgado: Extracted to: https://github.com/tasitlabs/TasitSDK/issues/372

pcowgill: Resolving this thread so we can continue the discussion in that new issue

packages/tasit-action/src/contract/Action.js
pcowgill: Ah right, because `GnosisSafe.executeTransaction` is an action-builder, not something that sends the action. I forgot that when reading it - it makes more sense now.

pcowgill: Or wait, maybe what I just said is wrong

pcowgill: I guess what I'm asking is this: At a high level, what is it about `GnosisSafe.executeTransaction` that prevents it from having a rawTx ready to go right away?

packages/tasit-action/src/contract/Contract.js
pcowgill: Yep, it is somewhat related to that.

Potentially before we close this thread, should we rename the setWallet and getWallet functions to setAccount and getAccount?

Or maybe `setSenderAddress` and `getSenderAddress` in case we want to set it to a contract-based account?

packages/tasit-contract-based-account/src/GnosisSafe.js
pcowgill: Sounds good - thanks

pcowgill: Resolving this

packages/tasit-action/src/contract/Action.js
marcelomorgado: > I guess what I'm asking is this: At a high level, what is it about `GnosisSafe.executeTransaction` that prevents it from having a rawTx ready to go right away?

The `[#prepareTransaction](https://github.com/tasitlabs/TasitSDK/blob/4111170669afb4447cb95315f45fd4ecd1fa02ec/packages/tasit-contract-based-account/src/GnosisSafe.js#L157)`.
That function has some async calls needed to create a `safe` transaction (e.g.: GnosisSafeContract.nonce (L199) and GnosisSafeContract.getTransactionHash (L201)).
These async calls are nested and needed to create the transaction that we want to be executed by the `GnosisSafeContract`.
My intention is to have the `GnosisSafe` encapsulating all that is possible, transparent to the user.

packages/tasit-action/src/contract/Contract.js
marcelomorgado: Oh, ok sure. I've thought that we'll want to rename it when the `Account` be !== `ethers.Wallet`.
Extracted to: https://github.com/tasitlabs/TasitSDK/issues/374

pcowgill: Ah ok. Nah, I think we want to pick our own names that we think are best even when we're using something directly from ethers.

pcowgill: Resolving this

packages/tasit-action/src/contract/Action.js
pcowgill: transparent to the user or opaque to the user?

marcelomorgado: :man_facepalming: I don't know why I've used the "transparent" word. Opaque was what I've meant.

Issue #329, opened 2019-04-18
Solve non-deterministic tests


marcelomorgado:
Deal with already skipped test cases plus:

TasitAction.Contract > `should emit error event when block reorganization occurs - tx confirmed twice` ([refs](https://github.com/tasitlabs/TasitSDK/pull/328#issuecomment-484576218))

marcelomorgado:
Some test cases still disabled (skipped) and some confirmations/blocks count stills as non-deterministic.
I think that all cases could be solved when these issues were addressed:
- https://github.com/trufflesuite/ganache-core/issues/248
- https://github.com/ethers-io/ethers.js/issues/393
- https://stackoverflow.com/a/50785284

pcowgill:
@marcelomorgado Makes sense. Thanks for the update!

marcelomorgado:
Recent failures to be addressed:

[TasitAction.Contract > Transactions (Actions) Subscription > should be able to listen to an event before sending](https://circleci.com/gh/tasitlabs/TasitSDK/601)
Move from `to.equal` to `to.be.at.least`

[TasitAction.Contract > Transactions (Actions) Subscription > should emit error event when block reorganization occurs - tx confirmed twice](https://circleci.com/gh/tasitlabs/TasitSDK/602)
Skip for now since no call was made to the listener.

marcelomorgado:
Since PR https://github.com/tasitlabs/TasitSDK/pull/386 we haven't non-deterministic test cases anymore.
We can close this or wait for more test rounds before make sure that issues were solved.

pcowgill:
Okay, great!! Thanks. I'll set a reminder to close this in two days if we don't run into any intermittent issues.

pcowgill:
The test suite passed in this PR https://github.com/tasitlabs/TasitSDK/pull/389 , so that's good

pcowgill:
@marcelomorgado This merge into develop didn't have the tests pass:

https://circleci.com/gh/tasitlabs/TasitSDK/655

```
tasit-action:   1) TasitAction.Contract
tasit-action:        Contract Events Subscription
tasit-action:          contract should call error listener after timeout:
tasit-action:      Error: Timeout of 8000ms exceeded. For async tests and hooks, ensure "done()" is called; if returning a Promise, ensure it resolves. (/home/circleci/tasit-sdk/packages/tasit-action/src/contract/Contract.test.js)
```

marcelomorgado:
[...]

Issue #367, opened 2019-05-01
Contract event listeners aren't working properly


marcelomorgado:
Same test case is working with `action.once('confirmation')` + `action.on('error')` listeners but it doesn't with `contract.once('Event')` + `contract.on('error')`.

Investigate why.

See more:
Decentraland.test.js > Using an ephemeral wallet funded by faucets > should buy an estate

pcowgill:
@marcelomorgado Good to know - thanks for flagging this

marcelomorgado:
Seems that issue is related to [parallelism event call](https://github.com/tasitlabs/TasitSDK/issues/380) that was fixed on `Action` only. After fixing that on `Contract` too (Pulling it up to `Subscription` class) tests were passed. The fixes were made on a coming PR.

pcowgill:
@marcelomorgado Great, thanks!

PR #369, opened 2019-05-01
Test cases from waiting to event driven style

### LINKED ISSUE #273: Decentraland + Gnosis Safe: Create test cases for a failed flow
That flow isn't working because of the highlighted situation:

I) Pre conditions:
-----------------
1. Gnosis safe wallet is funded with enough mana to buy some parcels/estates.
2. Gnosis safe wallet is funded with enough ethers to transfer to ephemeral account.

II) On boarding:
--------------
1. Ephemeral account will be funded with some ethers to pay for gas (from Gnosis safe wallet).
2. Gnosis safe approves ephemeral account to spend their mana tokens.
3. Ephemeral account approves Marketplace to spend their mana tokens (actually tokens from Gnosis Safe). **Not sure if this allowance-of-allowance will work.**

III) Buy operation:
----------------
1. Ephemeral account executes an open order from Marketplace
2. The asset bought will go to the ephemeral account balance.

A test case should be created to cover this case.

The step II.3 doesn't work. The marketplace executes `erc20.transferFrom(msg.sender, ...)` so the ephemeral account should have balance.

**MANAToken**
```solidity
// msg.sender here is the marketplace contract

 function transferFrom(address _from, address _to, uint256 _value) returns (bool) {
   var _allowance = allowed[_from][msg.sender];

   // Check is not needed because sub(_allowance, _value) will already throw if this condition is not met
   // require (_value <= _allowance);

   balances[_to] = balances[_to].add(_value);
   balances[_from] = balances[_from].sub(_value);
   allowed[_from][msg.sender] = _allowance.sub(_value);
   Transfer(_from, _to, _value);
   return true;
 }
```

marcelomorgado:
### Issue link

https://github.com/tasitlabs/TasitSDK/issues/329
https://github.com/tasitlabs/TasitSDK/issues/273
https://github.com/tasitlabs/TasitSDK/issues/219
https://github.com/tasitlabs/TasitSDK/pull/354#discussion_r278745008

### Auto-close the issue?

Closes #273

### Types of changes

### Notes

This change should be extended to all test cases of the project?


packages/tasit-action/src/contract/Action.js
```diff
@@ -171,7 +169,9 @@ export class Action extends Subscription {
           },
         };
 
+        // Note: Unsubscription should be done after the user's listener function be called
```
pcowgill:
```suggestion
        // Note: Unsubscribing should be done after the user's listener function is called
```

packages/tasit-action/src/contract/Action.js
```diff
@@ -180,16 +180,29 @@ export class Action extends Subscription {
...
-      ethersListener = async blockNumber => {
-        await baseEthersListener(blockNumber);
-        this.off(eventName);
-      };
-    } else {
-      ethersListener = baseEthersListener;
-    }
+    // Note:
+    // On the development env (using ganache-cli)
+    // Blocks are being mined simultaneously and generating a sort of unexpected behaviors like:
```
pcowgill: Maybe if we pick a very, very short but nonzero block time it would help?

packages/tasit-action/src/contract/Action.js
```diff
@@ -180,16 +180,29 @@ export class Action extends Subscription {
...
+    // - sequential blocks giving same confirmation to a transaction
+    // - false-positive reorg event emission
+    // - collaborating for tests non-determinism
+    //
+    // Tech debt:
+    // See if there is another way to avoid these problems, if not
+    // this solution should be improved with a state structure identifying state per event
+    //
+    // Question:
+    // Is possible that behavior (listener concurrency calls for the same event) be desirable?
```
pcowgill:
```suggestion
    // Is it possible that that behavior (listener concurrent calls for the same event) is desirable?
```

packages/tasit-action/src/contract/Contract.js
```diff
@@ -117,9 +117,9 @@ export class Contract extends Subscription {
           },
         };
 
-        if (once) this.off(eventName);
-
+        // Note: Unsubscription should be done after the user's listener function be called
```
pcowgill:
```suggestion
        // Note: Unsubscripting should be done after the user's listener function is called
```

packages/tasit-action/src/contract/Contract.test.js
```diff
@@ -147,7 +147,7 @@ describe("TasitAction.Contract", () => {
       });
 
       it("and Action error event on action error", async () => {
-        const contractErrorListener = sinon.fake(() => {
+        const contractErrorListener = sinon.fake(error => {
           sampleContract.off("error");
```
pcowgill: We probably don't actually need to remove the error listener in a test like this, right? Is it here because of the instantaneous block times?

packages/tasit-action/src/contract/Contract.test.js
```diff
@@ -175,24 +175,19 @@ describe("TasitAction.Contract", () => {
...
 
         sampleContract.on("error", errorListener);
         sampleContract.on("ValueChanged", eventListener);
 
-        action = sampleContract.setValue("hello world");
-        await action.send();
-
-        await action.waitForOneConfirmation();
-
-        await mineBlocks(provider, 4);
```
pcowgill: It's nice that there are no awaits in here now

packages/tasit-action/src/contract/Contract.test.js
```diff
@@ -541,7 +534,35 @@ describe("TasitAction.Contract", () => {
 
       await mineBlocks(provider, 2);
 
-      expect(confirmationListener.callCount).to.equal(1);
+      // Non-determinitic
+      expect(confirmationListener.callCount).to.be.at.least(1);
+      expect(errorListener.called).to.be.false;
+    });
+
+    it("once listener should be unsubscribed only after user listener function be called", async () => {
```
pcowgill:
```suggestion
    it("once listener should be unsubscribed only after user listener function was called", async () => {
```

packages/tasit-action/src/contract/Contract.test.js
```diff
@@ -541,7 +534,35 @@ describe("TasitAction.Contract", () => {
...
+      });
+
+      const confirmationListener = sinon.fake(message => {
+        console.log(message);
+      });
+
+      action.on("error", errorListener);
+      action.once("confirmation", confirmationListener);
+
+      // Forcing internal (block) listener be called before the transaction be sent
```
pcowgill:
```suggestion
      // Forcing internal (block) listener to be called before the transaction is sent
```

packages/tasit-action/src/contract/Action.js
pcowgill: @marcelomorgado This captures your intended meaning, right?

packages/tasit-contracts/src/scripts/populateDecentralandContracts.js
```diff
@@ -252,16 +252,16 @@ const createEstate = async estate => {
...
       action.unsubscribe();
       reject();
     });
 
-    estateContract.on("error", message => {
-      const { error } = message;
-      console.log(error.message);
+    estateContract.on("error", error => {
+      const { message } = error;
+      console.log(message);
```
pcowgill: Should it be `console.warn`?

packages/tasit-sdk/src/Decentraland.test.js
```diff
@@ -136,111 +136,176 @@ describe("Decentraland", () => {
...
+              await expectExactTokenBalances(estate, [ephemeralAddress], [1]);
+              done();
+            };
+
+            const errorListener = error => {
+              marketplace.off("error");
+              done(error);
+            };
+
+            // Note: These listener aren't working properly
```
pcowgill:
```suggestion
            // Note: These listeners aren't working properly
```

packages/tasit-sdk/src/Decentraland.test.js
```diff
@@ -136,111 +136,176 @@ describe("Decentraland", () => {
...
+
+            const orderSuccessfulListener = async message => {
+              const { data } = message;
+              const { args } = data;
+              const { buyer } = args;
+
+              expect(buyer).to.equal(ephemeralAddress);
+              await expectExactTokenBalances(estate, [ephemeralAddress], [1]);
+
+              done();
```
pcowgill: If this were working properly and it wasn't commented out where this was added, we wouldn't actually want the test to race for two `done`s, right? Instead we would split this out into two tests or pick just one style? Or not call done until both were true?

packages/tasit-sdk/src/Decentraland.test.js
```diff
@@ -136,111 +136,176 @@ describe("Decentraland", () => {
...
+              await expectExactTokenBalances(land, [ephemeralAddress], [1]);
+              done();
+            };
+
+            const errorListener = error => {
+              marketplace.off("error");
+              done(error);
+            };
+
+            // Note: These listener aren't working properly
```
pcowgill:
```suggestion
            // Note: These listeners aren't working properly
```

packages/tasit-sdk/src/Decentraland.test.js
```diff
@@ -249,393 +314,657 @@ describe("Decentraland", () => {
...
-            seller,
-            priceInWei,
-            expiresAt,
-          } = estateForSale;
+            action.send();
+          })();
+        });
 
-          await checkAsset(estate, mana, estateForSale, ephemeralAddress);
+        beforeEach("onboarding - funding ephemeral wallet with MANA", done => {
```
pcowgill: Out of curiosity, why not have multiple things happen in a single `beforeEach`?

packages/tasit-sdk/src/Decentraland.test.js
```diff
@@ -249,393 +314,657 @@ describe("Decentraland", () => {
...
+                ephemeralAddress,
+                MARKETPLACE_ADDRESS
+              );
+
+              expect(`${allowance}`).to.equal(`${manaAmountForShopping}`);
+              done();
+            };
+
+            const errorListener = error => {
+              action.off("error");
```
pcowgill: Any reason this is needed? Maybe as part of test cleanup all listeners should be removed?

packages/tasit-sdk/src/Decentraland.test.js
```diff
@@ -249,393 +314,657 @@ describe("Decentraland", () => {
...
-
-          const buyerAddress = GNOSIS_SAFE_ADDRESS;
-          await checkAsset(estate, mana, estateForSale, buyerAddress);
-
-          await expectExactTokenBalances(estate, [GNOSIS_SAFE_ADDRESS], [0]);
+          const toAddress = ephemeralAddress;
 
-          // Gnosis Safe should execute an open order
+          // Funding ephemeral account with some ethers to pay for gas
           // TODO: ephemeralWallet should broadcast this action
```
pcowgill: That's an interesting idea - curious for your rationale here

pcowgill: Then it's more like a Tasit app clicking a "fund myself" button?

packages/tasit-sdk/src/Decentraland.test.js
```diff
@@ -249,393 +314,657 @@ describe("Decentraland", () => {
...
+            const onFailed = message => {
+              const { data } = message;
+              const { args } = data;
+              const { txHash } = args;
+              done(new Error(`ExecutionFailed event emitted`));
+            };
+
+            const confirmationListener = async message => {
+              // WIP: Not working because of gas issue on Marketplace.safeExecuteOrder() call
+              //await expectExactTokenBalances(estate, [GNOSIS_SAFE_ADDRESS], [1]);
```
pcowgill: So in other words, we want to be expecting that it's 1, but instead we have to expect that it's 0 until we figure out the gas thing? If so, we can say that even more explicitly in the comments

packages/tasit-sdk/src/Decentraland.test.js
```diff
@@ -249,393 +314,657 @@ describe("Decentraland", () => {
...
+                );
+                expect(`${marketplaceAllowance}`).to.equal(
+                  `${manaAmountForShopping}`
+                );
+                done();
+              };
+
+              const errorListener = error => {
+                action.off("error");
+                done(error);
```
pcowgill: Isn't this the test `onboarding - approving Marketplace to spend ephemeral account's MANA` where for now we want to assert that it won't work? So should we be doing an expectation here that the error listener is triggered?

pcowgill: Oh never mind, I see that that is done below - we're in the beforeEach here, and we're using done() syntax there. Got it.

packages/tasit-sdk/src/Decentraland.test.js
```diff
@@ -249,393 +314,657 @@ describe("Decentraland", () => {
...
+              `${priceInWei}`,
+              `${fingerprint}`
+            );
+
+            const confirmationListener = async message => {
+              // Allowance-of-allowance doesn't work
+              // See more: https://github.com/tasitlabs/TasitSDK/issues/273
+              //await expectExactTokenBalances(estate, [ephemeralAddress], [1]);
+              await expectExactTokenBalances(estate, [ephemeralAddress], [0]);
+              done();
```
pcowgill: Is there any reason to call `done()` here if we know allowance-of-allowance doesn't work and we won't get to this path? Should we make it `done` with a new error here for now so the test won't pass if it gets here?

packages/tasit-sdk/src/Decentraland.test.js
```diff
@@ -249,393 +314,657 @@ describe("Decentraland", () => {
...
+              action.off("error");
+
+              // Note: once('confirmation') isn't unsubscribing after error emission
+              // See more: https://github.com/tasitlabs/TasitSDK/issues/368
+              action.off("confirmation");
+
+              // Allowance-of-allowance doesn't work
+              // See more: https://github.com/tasitlabs/TasitSDK/issues/273
+              //done(error);
+              done();
```
pcowgill: We can probably find a clever test helper to use like "assert error listener triggered", etc.

packages/tasit-action/src/contract/Action.js
marcelomorgado: I've tried that but not with extremely low block time. I'll make a try.

packages/tasit-action/src/contract/Action.js
marcelomorgado: Yep.

packages/tasit-contracts/src/scripts/populateDecentralandContracts.js
marcelomorgado: Yes, sounds good.

packages/tasit-sdk/src/Decentraland.test.js
marcelomorgado: Are you saying about both `orderSuccessfulListener` and `confirmationListener` on the same test case? If so, we are using only the `confirmationListener`. The intention is to use the `OrderSucessful` event to trigger listener but it isn't working right now (https://github.com/tasitlabs/TasitSDK/issues/367). Because of that, I've added the `confirmationListener` for now.

packages/tasit-sdk/src/Decentraland.test.js
marcelomorgado: Yes, it's related to https://github.com/tasitlabs/tasit/issues/151#issuecomment-474908828 discussion.
I've tried to avoid "callback hell" since these actions are nested.

packages/tasit-sdk/src/Decentraland.test.js
marcelomorgado: Yes, we can move to share `action` var between all test cases and having a cleanup (same as Contract.test.js), but I like the way that it is because reflects better how it's should be used on a real case. Thoughts?

packages/tasit-sdk/src/Decentraland.test.js
marcelomorgado: Yes.

packages/tasit-sdk/src/Decentraland.test.js
marcelomorgado: Yes. Ok, I'll be more explicitly.

packages/tasit-sdk/src/Decentraland.test.js
marcelomorgado: I've understood that we want to force them pass (refs: https://github.com/tasitlabs/TasitSDK/issues/273#issuecomment-473977995).
Maybe be more explicitly on the comments? Related to https://github.com/tasitlabs/TasitSDK/pull/369#discussion_r280225416

packages/tasit-sdk/src/Decentraland.test.js
marcelomorgado: Hmm, I'll take a look.

pcowgill: Great - thanks

packages/tasit-action/src/contract/Action.js
pcowgill: Okay - can you please include this change in the next PR?

packages/tasit-sdk/src/Decentraland.test.js
pcowgill: Yeah, I'm saying the way it's written right now it's not clear if you'd use both listeners in the same test case, because we kind of are now even though one is commented out.

I think my main preference would be a unit test somewhere that confirms that either method would work in two separate tests. And then in our functional tests throughout the rest of the code, we just use events like you're suggesting.

Maybe include a comment in the code to signal this intent to only use events here in the future?

packages/tasit-sdk/src/Decentraland.test.js
pcowgill: Ah ok. Do these `beforeEach`'s happen sequentially or in parallel.

The other way to avoid "callback hell" in a single `beforeEach` would be to ship marketplace approval and mana funding in parallel, have a separate listener for each that sets local state for whether it has it's response yet, and each time when setting state check if the other one is already done, and if so finish the whole before each.

Is the only callback hell here that we wouldn't want to ship that double action only once the single action of receiving eth had completed?

```js
// buildManaAction here
// buildEthAction here
// buildMarketplaceAction here

functionThatSetsManaDone {
  if (marketplaceDone) {
    functionThatHandlesMarketplaceAndManaDone()
  }
}

functionThatShipsManaAndMarketplaceApproval {
  manaAction.send()
  marketplaceAction.send()
}

ethAction.on('confirmation', functionThatShipsManaAndMarketplaceApproval)
manaAction.on('confirmation', functionThatSetsManaDone)
marketplaceAction.on('confirmation', functionThatSetsMarketplaceDone)

ethAction.send()
```

I guess it is a little confusing to reason about, but as long as we extract the function and define it ahead of time, we're not actually defining functions in a nested way that creates lots of indentation, although there is nesting of program flow.

And we could of course listen for events rather than confirmations above

packages/tasit-sdk/src/Decentraland.test.js
pcowgill: Hmm, so you're saying the user in reality should remove an error listener the moment they get an error? Why? What if there were some second relevant error?

packages/tasit-sdk/src/Decentraland.test.js
pcowgill: I guess if a user in the Tasit app clicked fund myself, it would still work just as well if it deep-linked into Gnosis Safe app and then the Gnosis Safe app broadcast it, right?

packages/tasit-sdk/src/Decentraland.test.js
pcowgill: Great, thanks!

packages/tasit-sdk/src/Decentraland.test.js
pcowgill: Yeah, we do want to force them to pass. But if we know ahead of time that the code will only flow through one path (the error path) and that's what we want to assert/expect is happening, couldn't we just have one done in the listener for the error path and not here (in the conf listener)?

packages/tasit-sdk/src/Decentraland.test.js
marcelomorgado: Good point. This is the way that is working now. So, maybe we what that the error listener works always on the "`once` mode"?

pcowgill: Hmm wait, I'm still confused - I think we may disagree?

I think users will usually want to listen to errors forever, with some exceptions.

Are you saying that users will usually only want to listen to the first error?

packages/tasit-sdk/src/Decentraland.test.js
marcelomorgado: I think so, that test is using the gnosisOwner to call the gnosisSafe contract. So it's like to **not** having the `ephemeral` account as a signer and calling the gnosisSafeApp thru deep-link.

packages/tasit-sdk/src/Decentraland.test.js
marcelomorgado: Ohh, got it. Yes, I agree. I'll remove the `done` from confirmationListener.

pcowgill: Awesome, thanks!

packages/tasit-sdk/src/Decentraland.test.js
pcowgill: Yep. So depending on the scope of this particular test, either option is a sensible choice.

Maybe add something from this GitHub comment thread to explain that in addition to the current TODO comment in the code?

packages/tasit-sdk/src/Decentraland.test.js
marcelomorgado: I think we maybe will have both situations. Examples:
- Keep `error` listener running and turn it off when the action was successful;
- Turn the `error` listener off if an error occurs (for some reason).
Yes, not make sense unsubscribe after an error emission.


So, answering your question again:

> Any reason this is needed? Maybe as part of test cleanup all listeners should be removed?

You call, I don't have a strong opinion but I like to not have to use global var and having a complete isolated test case.

packages/tasit-sdk/src/Decentraland.test.js
marcelomorgado: Ok, I'll do that.

packages/tasit-sdk/src/Decentraland.test.js
pcowgill: I had a little confusion stemming from picturing how users would use this feature vs. how we want to architect the test suite.

I agree that not having to use a global var in the test suite is nice. So we don't need to create any test cleanup setup that handles cleanup for listeners from all tests.

And I agree that users or our test suite might want to do like you described in both situations above.

What's the problem that we would run into if we kept the error listener subscribed unnecessarily here? Just memory leak at scale? Or intermittent test failures?

pcowgill: Long story short - we can keep this line of code as-is, but I'm still curious about the last question for future reference.

packages/tasit-action/src/contract/Action.js
marcelomorgado: I've tried these blockTimes: 0.001, 0.01 and 0.1. The lowest values increased the occurrences of the parallel calls.
I guess that could be related to auto block generation for each transaction by `ganache-cli` and maybe this feature could help (https://github.com/trufflesuite/ganache-core/issues/248).
I've open this issue to track that: https://github.com/tasitlabs/TasitSDK/issues/380

packages/tasit-action/src/contract/Contract.test.js
marcelomorgado: Right. The `Contract.test.js` differently from `Decentraland.test.js` is cleaning listener between test cases. That will be addressed on a PR that I'm working on.

packages/tasit-contracts/src/scripts/populateDecentralandContracts.js
marcelomorgado: Solve on https://github.com/tasitlabs/TasitSDK/pull/379

pcowgill: Great, thanks! Resolving this.

packages/tasit-action/src/contract/Contract.test.js
pcowgill: Okay, great. Thanks for the info and update.

packages/tasit-action/src/contract/Action.js
pcowgill: Yep, agreed that that is the most likely cause. I agree that that Ganache feature could help. I added a comment on that issue.

Is the non-determinism mostly for tests where the listener is added before or after sending the tx? Or is it a little of both?

We're sure it's not just getting called again with another confirmation for each instantaneous block, right? If so, we could probably disable that in our code and check number of confirmations on each listener call. Or something like that.

pcowgill: Feel free to move my comment over to that issue instead

pcowgill: If we move it there, we can resolve this thread. Thanks for testing those block times!

marcelomorgado: Comment extracted to issue. Resolving this.

packages/tasit-sdk/src/Decentraland.test.js
marcelomorgado: Extracted to: https://github.com/tasitlabs/TasitSDK/issues/381

packages/tasit-sdk/src/Decentraland.test.js
marcelomorgado: > What's the problem that we would run into if we kept the error listener subscribed unnecessarily here? Just memory leak at scale? Or intermittent test failures?

Mocha can be stuck (failing test suite by timeout), not sure but non-determinism can occur since we are using the same provider for all tests on the same test suite.

pcowgill: That makes sense. Thanks. Resolving this.

packages/tasit-sdk/src/Decentraland.test.js
pcowgill: Thanks. Resolving this.

packages/tasit-sdk/src/Decentraland.test.js
marcelomorgado: Extracted to: https://github.com/tasitlabs/TasitSDK/issues/382

packages/tasit-sdk/src/Decentraland.test.js
marcelomorgado: Done on https://github.com/tasitlabs/TasitSDK/pull/383

packages/tasit-sdk/src/Decentraland.test.js
marcelomorgado: Could I resolve this?

packages/tasit-sdk/src/Decentraland.test.js
marcelomorgado: Done on https://github.com/tasitlabs/TasitSDK/pull/383

packages/tasit-sdk/src/Decentraland.test.js
marcelomorgado: I've splitted the test case and now we have one using contract events and another using action events.

packages/tasit-action/src/contract/Contract.test.js
marcelomorgado: Could I close this?

packages/tasit-action/src/contract/Action.js
marcelomorgado: Done on https://github.com/tasitlabs/TasitSDK/pull/383 (Now this comment lives on Subscription class)

pcowgill: Thanks! Resolving this.

packages/tasit-action/src/contract/Contract.test.js
pcowgill: Yep. Thanks!

packages/tasit-sdk/src/Decentraland.test.js
pcowgill: Great, thanks!! Resolving this

packages/tasit-sdk/src/Decentraland.test.js
pcowgill: Thanks! Resolving this

packages/tasit-sdk/src/Decentraland.test.js
pcowgill: Yep, I'm resolving it now

packages/tasit-sdk/src/Decentraland.test.js
pcowgill: Great, thanks! Resolving this

PR #379, opened 2019-05-03
ESLint

marcelomorgado:
### Issue link

#361 
#329
#219
#237 

### Auto-close the issue?

Closes #361 
Closes item 5 of #237

### Types of changes

Technical debt (a code change that doesn't fix a bug or add a feature but makes something clearer for devs)

### Notes

Lint scripts were moved to root otherwise, we'll have to install linter on all packages (Based on the discussion: https://github.com/lerna/lerna/issues/2005).
I've realized that our scripts weren't using the `prettier` bin from root repo but installing it via `npx`.

PR #386, opened 2019-05-07
Event timeouts

### LINKED ISSUE #329: Solve non-deterministic tests

### LINKED ISSUE #384: Make sure that events timeout are working
Our only test case that is testing that is being skipped because it seems non-deterministic (because of lack of fake timer + async + mocha). We can move it to sync/event style to make sure if the timeout is working.

marcelomorgado:
### Issue link

#384 
#329
https://github.com/tasitlabs/TasitSDK/pull/379#discussion_r280867479
https://github.com/tasitlabs/TasitSDK/pull/379#discussion_r280864116

### Auto-close the issue?

Closes #384 

### Types of changes

### DIFF
--- a/packages/tasit-action/src/contract/Contract.test.js
+++ b/packages/tasit-action/src/contract/Contract.test.js
@@ -4,7 +4,6 @@ import ProviderFactory from "../ProviderFactory";
 import {
   accounts,
   mineBlocks,
-  wait,
   createSnapshot,
   revertFromSnapshot,
 } from "../testHelpers/helpers";
--- a/packages/tasit-action/src/contract/Contract.test.js
+++ b/packages/tasit-action/src/contract/Contract.test.js
@@ -334,45 +334,30 @@ describe("TasitAction.Contract", () => {
       expect(errorListener.called).to.be.false;
     });
 
-    // Non-deterministic test case
-    it.skip("should call error listener after timeout", async () => {
+    it("action should call error listener after timeout", done => {
       action = sampleContract.setValue("hello world");
       action.setEventsTimeout(100);
 
       const errorListener = sinon.fake(error => {
-        expect(error.eventName).to.equal("confirmation");
-        expect(error.message).to.equal("Event confirmation reached timeout.");
+        const { eventName, message } = error;
+        expect(eventName).to.equal("confirmation");
+        expect(message).to.equal("Event confirmation reached timeout.");
+        expect(action.subscribedEventNames()).to.deep.equal([
+          "error",
+          "confirmation",
+        ]);
         action.off("error");
+        action.off("confirmation");
+        done();
       });
 
-      const confirmationListener = sinon.fake(() => {
-        action.off("confirmation");
-      });
+      const confirmationListener = sinon.fake(() => {});
 
       action.on("confirmation", confirmationListener);
-
       action.on("error", errorListener);
 
-      await action.send();
-      await action.waitForOneConfirmation();
-
-      await mineBlocks(provider, 2);
-
-      // TODO: Use fake timer when Sinon/Lolex supports it.
-      // See more:
-      //  https://github.com/sinonjs/sinon/issues/1739
-      //  https://github.com/sinonjs/lolex/issues/114
-      //  https://stackoverflow.com/a/50785284
-      await wait(action.getEventsTimeout() * 5);
-
-      await mineBlocks(provider, 2);
-
-      expect(errorListener.callCount).to.equal(1);
-      expect(confirmationListener.callCount).to.equal(1);
-      expect(action.subscribedEventNames()).to.deep.equal([
-        "error",
-        "confirmation",
-      ]);
+      // Note: No time is necessary because no additional block will be mined
+      action.send();
     });
 
     it("subscription should have one listener per event", async () => {
@@ -622,6 +607,32 @@ describe("TasitAction.Contract", () => {
       sampleContract.on("error", errorListener);
       sampleContract.on("ValueChanged", valueChangedListener);
 
+      action.send();
+    });
+
+    it("contract should call error listener after timeout", done => {
+      sampleContract.setEventsTimeout(100);
+      action = sampleContract.setValue("hello world");
+
+      const errorListener = sinon.fake(error => {
+        const { eventName, message } = error;
+        expect(eventName).to.equal("ValueChanged");
+        expect(message).to.equal("Event ValueChanged reached timeout.");
+        expect(sampleContract.subscribedEventNames()).to.deep.equal([
+          "ValueChanged",
+          "error",
+        ]);
+        sampleContract.off("error");
+        sampleContract.off("ValueChanged");
+        done();
+      });
+
+      const confirmationListener = sinon.fake(() => {});
+
+      sampleContract.on("ValueChanged", confirmationListener);
+      sampleContract.on("error", errorListener);
+
+      // Note: No time is necessary because no additional block will be mined
       action.send();
     });
 
--- a/packages/tasit-action/src/contract/Subscription.js
+++ b/packages/tasit-action/src/contract/Subscription.js
@@ -1,8 +1,15 @@
+import ConfigLoader from "../ConfigLoader";
+
 export class Subscription {
   #ethersEventEmitter;
   #eventListeners = new Map();
+  #timeout;
 
   constructor(eventEmitter) {
+    const { events } = ConfigLoader.getConfig();
+    const { timeout } = events;
+
+    this.#timeout = timeout;
     this.#ethersEventEmitter = eventEmitter;
   }
 
@@ -27,6 +34,14 @@ export class Subscription {
       );
     }
     this.#eventListeners.delete(eventName);
+  };
+
+  getEventsTimeout = () => {
+    return this.#timeout;
+  };
+
+  setEventsTimeout = timeout => {
+    this.#timeout = timeout;
   };
 
   // TODO: Make protected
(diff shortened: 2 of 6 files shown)


pcowgill:
I updated the description of the PR to remove the part about closing the non-deterministic tests issue automatically based on my comment over in that issue

packages/tasit-action/src/contract/Contract.test.js
```diff
@@ -334,45 +333,30 @@ describe("TasitAction.Contract", () => {
...
-
-      await mineBlocks(provider, 2);
-
-      expect(errorListener.callCount).to.equal(1);
-      expect(confirmationListener.callCount).to.equal(1);
-      expect(action.subscribedEventNames()).to.deep.equal([
-        "error",
-        "confirmation",
-      ]);
+      // Note: No time is necessary because no additional block will be mined
```
pcowgill: But if it’s calling it after a timeout, I don’t totally understand what this means here. Do you mean no block mining is necessary, as long as the event timeout is shorter than the test timeout?

packages/tasit-action/src/contract/Subscription.js
```diff
@@ -29,6 +36,14 @@ export class Subscription {
     this.#eventListeners.delete(eventName);
   };
 
+  getEventsTimeout = () => {
+    return this.#timeout;
+  };
+
+  setEventsTimeout = timeout => {
```
pcowgill: Should we call it something other than events to indicate that it can relate to confirmations too and not just “contract events”?

packages/tasit-sdk/src/Decentraland.test.js
```diff
@@ -723,16 +714,67 @@ describe("Decentraland", () => {
...
+              // const { data } = message;
+              // const { args } = data;
+              // const { txHash } = args;
+              //done(new Error(`ExecutionFailed event emitted`));
+              done();
+            };
+
+            const errorListener = error => {
+              action.off("error");
+              done(error);
```
pcowgill: Maybe note here that because ExecutionFailed is a custom contract event, we don’t pass it through as an error event yet. Otherwise we’d risk failing the test here. We may eventually want to detect that specifically for the Safe and call it an error.

packages/tasit-action/src/contract/Contract.test.js
marcelomorgado: - Test case timeout (mocha) is 8 sec;
- Test case will wait until `done()` is called;
- Timeout error will be thrown if the time between blocks be > timeout;
That said, after first confirmation, since now the additional block is being called, the timeout error is being thrown.
Make sense?

packages/tasit-action/src/contract/Subscription.js
marcelomorgado: Ok, `setTimeout` sounds good?

packages/tasit-action/src/contract/Contract.test.js
marcelomorgado: > Timeout error is thrown in practice mostly if there’s a new block without a new confirmation for that tx

Timeout error is thrown if the time between our `ethers.js` wrapper listener triggers is greater than the timeout. (I'll put the code below). It will occur for block event ('confirmation') and contract event.

> But the lack of a block also works.

Yes, because we are using [`ethers.js` block event](https://docs.ethers.io/ethers.js/html/api-providers.html#event-types) for our `confirmation` event and without new block, the timeout will be reached.

> So the timeout time period happens after the block that is mined as a result of this send, right? (This is my understanding right now.)
Or did the time period happen already and end when this send happens?

Could clarify what you've meant with "as result of this send"?


Code from `Subcription` class

```javascript
const listener = async (...args) => {
  const eventListener = this.#eventListeners.get(eventName);
  const { isRunning } = eventListener;

  if (isRunning) {
    console.info(`Listener is already running`);
    return;
  }

  this.#decorateEventListener(eventName, {
    isRunning: true,
    lastEmissionTime: Date.now(),
  });

  this._clearEventTimerIfExists(eventName);

  const timer = setTimeout(() => {
    const eventListener = this.#eventListeners.get(eventName);
    const { lastEmissionTime } = eventListener;
    const currentTime = Date.now();
    const timedOut =
      currentTime - lastEmissionTime >= this.getEventsTimeout();

    if (timedOut) {
      this._emitErrorEventFromEventListener(
        new Error(`Event ${eventName} reached timeout.`),
        eventName
      );
    }
  }, this.getEventsTimeout());

  this._setEventTimer(eventName, timer);

  await baseListener(...args);
  this.#decorateEventListener(eventName, { isRunning: false });
};

const eventListener = { listener, isRunning: false };

this.#eventListeners.set(eventName, eventListener);

this.#ethersEventEmitter.on(this._toEthersEventName(eventName), listener);
```

packages/tasit-sdk/src/Decentraland.test.js
marcelomorgado: I didn't get it. Could you help me to understand?

pcowgill: I believe that with a done-styled test, if you call `done(error)` the test fails. Right now, there's no way the `errorListener` would be triggered, so there's no way that `done(error)` would be called making the test fail.

But the contract event here is called `ExecutionFailed`, so you could imagine adding some `GnosisSafe` custom code that catches that event and passes it through as an error so the that `errorListener` would be called here. If that happened faster than the event listener for `ExecutionFailed` was called, then this test would intermittently fail.

So it's not a problem now, but it could be in the future.

packages/tasit-action/src/contract/Contract.test.js
pcowgill: > It will occur for block event ('confirmation') and contract event.

But there's only a timeout for a contract event if we set one explicitly, right?

pcowgill: >> So the timeout time period happens after the block that is mined as a result of this send, right? (This is my understanding right now.)
Or did the time period happen already and end when this send happens?

> Could clarify what you've meant with "as result of this send"?

I mean that ethers.js is set to instantly mine a new block whenever a tx is broadcast. So the fact that a new block is mined is due to the `.send()`

packages/tasit-sdk/src/Decentraland.test.js
marcelomorgado: Ohh, got it now.
Shouldn't the test case fit with the current implementation and be changed if the `GnosisSafe` implementation changed as you've said?

pcowgill: Yep, that's a good argument! We can resolve this.

marcelomorgado: Resolving this.

packages/tasit-action/src/contract/Contract.test.js
marcelomorgado: > > It will occur for block event ('confirmation') and contract event.
> 
> But there's only a timeout for a contract event if we set one explicitly, right?

No, not really.

marcelomorgado: > > > So the timeout time period happens after the block that is mined as a result of this send, right? (This is my understanding right now.)
> > > Or did the time period happen already and end when this send happens?
> 
> > Could clarify what you've meant with "as result of this send"?
> 
> I mean that ethers.js is set to instantly mine a new block whenever a tx is broadcast. So the fact that a new block is mined is due to the `.send()`

> > > So the timeout time period happens after the block that is mined as a result of this send, right? (This is my understanding right now.)

Right.

pcowgill: > > > It will occur for block event ('confirmation') and contract event.
> > 
> > 
> > But there's only a timeout for a contract event if we set one explicitly, right?
> 
> No, not really.

In that case, as of right now we can't listen for `ExecutionFailed` forever for a contract. Is that correct?

pcowgill: > > > > So the timeout time period happens after the block that is mined as a result of this send, right? (This is my understanding right now.)
> > > > Or did the time period happen already and end when this send happens?
> > 
> > 
> > > Could clarify what you've meant with "as result of this send"?
> > 
> > 
> > I mean that ethers.js is set to instantly mine a new block whenever a tx is broadcast. So the fact that a new block is mined is due to the `.send()`
> 
> > > > So the timeout time period happens after the block that is mined as a result of this send, right? (This is my understanding right now.)
> 
> Right.

In that case, I think we can change the comment

pcowgill:
```suggestion
      // Note: Sending this triggers a block being mined,
      // and since no additional blocks will be mined,
      // the timeout will be reached before another confirmation occurs
```

marcelomorgado: > In that case, as of right now we can't listen for ExecutionFailed forever for a contract. Is that correct?

Yes, that's correct. Should I create an issue to having that listening forever as default?

pcowgill: Yeah, I think so, if you agree

marcelomorgado: Comment was changed on https://github.com/tasitlabs/TasitSDK/pull/392

marcelomorgado: > > In that case, as of right now we can't listen for ExecutionFailed forever for a contract. Is that correct?
> 
> Yes, that's correct. Should I create an issue to having that listening forever as default?

Extracted to: https://github.com/tasitlabs/TasitSDK/issues/393

pcowgill: > Comment was changed on #392

Awesome, thanks

pcowgill: > > Yeah, I think so, if you agree
> 
> Extracted to: #393

Perfect, resolving this

PR #392, opened 2019-05-09
Log ganache-cli output

pcowgill:
### Issue link

https://github.com/tasitlabs/TasitSDK/pull/386#discussion_r282224108

### Types of changes

### Notes

Still TBD how to have a version of ganache running that logs thing but still lets us automatically run migrations after that has been done for a single `npm start` command for local development


marcelomorgado:
We can pipe the `ganache-cli` process to a log file (ignored by git) instead of `/dev/null`. Make sense?

pcowgill:
> We can pipe the `ganache-cli` process to a log file (ignored by git) instead of `/dev/null`. Make sense?

Yeah, that's a good idea. And then the idea is you would do a `tail -f` for that file in another tab?

marcelomorgado:
> > We can pipe the `ganache-cli` process to a log file (ignored by git) instead of `/dev/null`. Make sense?
> 
> Yeah, that's a good idea. And then the idea is you would do a `tail -f` for that file in another tab?

Exactly.

marcelomorgado:
I'm not sure about `ganache-cli` `-v` option since that log too much, the user could add that if he wants.
With piping to a log, we can stay with existent scripts.

pcowgill:
> I'm not sure about `ganache-cli` `-v` option since that log too much, the user could add that if he wants.

Yeah, now that I've tried it locally I can see that that's definitely too much info

> With piping to a log, we can stay with existent scripts.

Yep, that's better.

marcelomorgado:
I've updated the PR adding the reference of the open thread that requested the comment change.

